PR #31, opened 2016-03-31
[DNM] kvprc: Operator -> Op

disksing:


ngaut:
LGTM

BusyJay:
LGTM, though I think Op maybe too short.

PR #141, opened 2017-03-03
pdpb: add region_id for GetRegionRequest

nolouch:
PTAL @siddontang @disksing @AndreMouche


disksing:
we already have `GetRegionByIDRequest` message.

nolouch:
I didn't found It , close

PR #143, opened 2017-03-11
*: add cycle.yml

coocood:
Fix CI

### DIFF
--- a/pkg/util/codec_test.go
+++ b/pkg/util/codec_test.go
@@ -2,31 +2,33 @@ package util
 
 import (
 	"bytes"
+	"testing"
 
-	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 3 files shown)


pkg/util/codec_test.go
```diff
@@ -2,31 +2,33 @@ package util
 
 import (
 	"bytes"
+	"testing"
 
-	"github.com/golang/protobuf/proto"
-	. "github.com/pingcap/check"
```
shenli: Why not use `check` package?

coocood: Adds more dependency.

coocood: We don't have this package in vendor.

shenli:
LGTM

PR #146, opened 2017-03-28
upgrade rust protobuf

zhangjinpeng1987:
@disksing @siddontang PTAL


BusyJay:
LGTM

overvenus:
Em~ Why update Cargo.lock instead of Cargo.toml? As a library, updating crates in Cargo.lock has no effect.

> For libraries the situation is somewhat different. A library is not only used by the library developers, but also any downstream consumers of the library. Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists). This is precisely because a library should not be deterministically recompiled for all users of the library.

http://doc.crates.io/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

PR #174, opened 2017-06-27
support delete range kv command

zhangjinpeng1987:


proto/errorpb.proto
```diff
@@ -29,6 +29,14 @@ message KeyNotInRegion {
     optional bytes end_key     = 4;
 }
 
+message RangeNotInRegion {
```
disksing: How about return _KeyNotInRegion_ if _start_key_ or _end_key_ is not in region?

zhangjinpeng1987: They are different.

zhangjinpeng1987:
@zimulala PTAL

zhangjinpeng1987:
@disksing PTAL

zimulala:
LGTM

disksing:
LGTM.

PR #176, opened 2017-06-28
*: include proto files

overvenus:
Explicitly include google and gogoproto proto files.

PTAL @siddontang @BusyJay @disksing


BusyJay:
Why?

overvenus:
For better structure and understanding.

BusyJay:
A simpler and cleaner way to do that is create a directory soft link. Although I don't see the point of doing it. Will the deleted files be created again once vendor is updated?

overvenus:
No, they are no longer be kept. https://github.com/pingcap/kvproto/pull/176/files#diff-b67911656ef5d18c4ae36cb6741b7965R5

PR #182, opened 2017-07-18
add java pom for build java dependency jar

ilovesoup:
Add pom files for java proto dependencies. They are published as jar and referred in other projects.
Note that gogo proto is introduced even though it's not used in Java anyway.
Need to make a gogo proto clean version of it.

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/kvproto/182)


pom.xml
```diff
@@ -0,0 +1,152 @@
...
+                  <id>copy-resources</id>
+                  <phase>validate</phase>
+                  <goals>
+                     <goal>copy-resources</goal>
+                  </goals>
+                  <configuration>
+                     <outputDirectory>${basedir}/proto/gogoproto</outputDirectory>
+                     <resources>
+                        <resource>
+                           <directory>${GOPATH}/src/github.com/gogo/protobuf/gogoproto</directory>
```
BusyJay: GOPATH?

pom.xml
```diff
@@ -0,0 +1,152 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+   <modelVersion>4.0.0</modelVersion>
+   <groupId>com.pingcap.tikv</groupId>
+   <artifactId>kvproto</artifactId>
+   <version>0.1.0</version>
```
BusyJay: Currently, rust use version `0.0.1`.

pom.xml
ilovesoup: this is equal to generate-rust.sh
gogoproto introduce special tags and options into proto files and this has to be compiled with   gogoproto dependencies.

README.md
```diff
@@ -6,5 +6,6 @@ Protocol buffer files for TiKV
 + Write your own protocol file in proto folder.
 + Run `make` to generate go and rust code. 
     We generate all go codes in pkg folder and rust in src folder.
++ Run `mvn clean install` to generate and install Java dependencies locally
```
ilovesoup: this is included in tikv-client as submodules. maven install is just for convenience of users who wants to use it this way.

pom.xml
BusyJay: You can use `${basedir}/_vendor/vendor/github.com/gogo/protobuf/gogoproto`, so java developer doesn't have to install go and setup GOPATH to vendor.

pom.xml
ilovesoup: changed.

pom.xml
ilovesoup: it seems tipb (tidb's proto) does not have gogoproto in their source code. I will withdraw this pr and waiting for two repo merge as discussed.

ilovesoup:
it seems tipb (tidb's proto) does not have gogoproto in their source code and can't dealt with the same way. I will withdraw this pr and waiting for two repo merge as discussed.

PR #188, opened 2017-08-16
add read bytes for storestats

UncP:
Signed-off-by: UncP <uncp.xu@gmail.com>


choleraehyq:
LGTM @disksing PTAL

disksing:
LGTM.

PR #189, opened 2017-09-04
coprocessor: introduce select_resp field to replace data field

tiancaiamao:
Add `SelectResponse` field to `coprocessor.Response`

@BusyJay @shenli


proto/coprocessor.proto
```diff
@@ -29,4 +29,5 @@ message Response {
     errorpb.Error region_error = 2;
     kvrpcpb.LockInfo locked = 3;
     string other_error = 4;
+    bytes select_resp = 5 [(gogoproto.customtype) = "github.com/pingcap/tipb/go-tipb.SelectResponse"];
```
BusyJay: So TiKV still needs to encode the responses to bytes?

tiancaiamao:
Test show no significant benefit for this change.

PR #190, opened 2017-09-05
kvrpcpb: init not_fill_cache in context

AndreMouche:
@tiancaiamao @zhangjinpeng1987 @BusyJay @hicqu PTAL


proto/kvrpcpb.proto
```diff
@@ -44,6 +44,7 @@ message Context {
     uint64 term = 5;
     CommandPri priority = 6;
     IsolationLevel isolation_level = 7;
+    bool no_cache = 8;
```
tiancaiamao: I think `no_cache` is not a good name, `not_update_cache` is more precise.

BusyJay: How about `fill_cache`? Or `cache_hint`?

tiancaiamao: Good idea! noun is better name than verb.

pkg/kvrpcpb/kvrpcpb.pb.go
```diff
@@ -230,6 +230,7 @@ type Context struct {
 	Term           uint64              `protobuf:"varint,5,opt,name=term,proto3" json:"term,omitempty"`
 	Priority       CommandPri          `protobuf:"varint,6,opt,name=priority,proto3,enum=kvrpcpb.CommandPri" json:"priority,omitempty"`
 	IsolationLevel IsolationLevel      `protobuf:"varint,7,opt,name=isolation_level,json=isolationLevel,proto3,enum=kvrpcpb.IsolationLevel" json:"isolation_level,omitempty"`
+	NotFillCache   bool                `protobuf:"varint,8,opt,name=not_fill_cache,json=notFillCache,proto3" json:"not_fill_cache,omitempty"`
```
disksing: I prefer `SkipFillCache`, double negation is evil.

tiancaiamao:
LGTM

disksing:
LGTM.

PR #195, opened 2017-09-20
debugpb: rename some fields

overvenus:
CC https://github.com/pingcap/tikv/pull/2299


proto/debugpb.proto
```diff
@@ -51,7 +51,9 @@ enum CF {
 
 message GetRequest {
     CF cf = 1;
-    bytes key_encoded = 2;
+    // An arbitrary key for retrieving value from TiKV.
+    // Note: Server uses it directly w/o any encoding.
+    bytes key = 2;
```
BusyJay: I think a `db` field is necessary.

proto/debugpb.proto
```diff
@@ -51,7 +51,9 @@ enum CF {
 
 message GetRequest {
     CF cf = 1;
```
BusyJay: Use string is more flexible.

proto/debugpb.proto
```diff
@@ -38,22 +42,20 @@ service Debug {
...
+    KV = 1;
+    RAFT = 2;
 }
 
 message GetRequest {
-    CF cf = 1;
-    // An arbitrary key for retrieving value from TiKV.
-    // Note: Server uses it directly w/o any encoding.
-    bytes key = 2;
+    bytes key = 1;
```
BusyJay: Please keep it in the order `db`, `cf` then `key`.

BusyJay:
LGTM

hicqu:
LGTM.

PR #205, opened 2017-10-18
debug: add MetaRegions RPC.

hicqu:
This RPC returns all meta regions' id. It can be used for running some debug commands in remot mode, such as `raft region info`.


hicqu:
@BusyJay @overvenus , PTAL.

hicqu:
@BusyJay @overvenus PTAL

BusyJay:
What is a meta region?

hicqu:
The name is unclear. In fact, this API returns regions have any keys between `REGION_META_MIN_KEY` and `REGION_META_MAX_KEY`.

PR #207, opened 2017-10-25
Add Import Service and UploadSST RPC

huachaohuang:
Allow a client to upload an SST file to TiKV.


huachaohuang:
Will move to another branch.

PR #214, opened 2017-11-28
coprocessor: add KeyRange in Response.

hicqu:
Add a KeyRange into coprocessor::Response in order to retry streaming requests.


tiancaiamao:
LGTM

PR #219, opened 2018-01-09
Add ImportSST Service

huachaohuang:
ImportSST provides a service to import a generated SST file to a region in TiKV.

In order to import an SST file to a region, the user should:
1. Retrieve the meta of the region according to the SST file's range.
2. Upload the SST file to the servers where the region's peers locate in.
3. Issue an ingest request to the region's leader with the SST file's metadata.

It's the user's responsibility to make sure that the SST file is uploaded to
the servers where the region's peers locate in, before issue the ingest
request to the region's leader. However, the region can be scheduled (so the
location of the region's peers will be changed) or split/merged (so the range
of the region will be changed), after the SST file is uploaded, but before
the SST file is ingested. So, the region's epoch is provided in the SST
file's metadata, to guarantee that the region's epoch must be the same
between the SST file is uploaded and ingested later.


proto/importpb.proto
```diff
@@ -0,0 +1,68 @@
...
+    bytes end = 2;
+}
+
+message SSTMeta {
+    bytes uuid = 1;
+    Range range = 2;
+    uint32 crc32 = 3;
+    uint64 length = 4;
+    string cf_name = 5;
+    uint64 region_id = 6;
```
zhangjinpeng1987: Should we consider the situation of region split or merge?

proto/importpb.proto
```diff
@@ -0,0 +1,68 @@
...
+    metapb.RegionEpoch region_epoch = 7;
+}
+
+message UploadRequest {
+    SSTMeta meta = 1;
+    bytes data = 2;
+}
+
+message UploadResponse {
```
zhangjinpeng1987: Also need a error filed?

proto/importpb.proto
huachaohuang: Yes, it is considered. The region's epoch will be checked before ingestion. So, if the region's epoch is changed after the SST file has been uploaded, the following ingestion will fail.

proto/importpb.proto
huachaohuang: The error will be returned as a GRPC error, and we don't have other special errors here, so no need to add an extra error field.

zhangjinpeng1987:
LGTM

zhangjinpeng1987:
@UncP PTAL

PR #223, opened 2018-01-23
add prefix key filter for tikv ctl

wentaoxu:
filter all the regions which size is larger than some limitation.


wentaoxu:
PTAL @zhangjinpeng1987

BusyJay:
Can this be implemented purely in tikv-ctl?

wentaoxu:
@BusyJay I'm afraid not for `DebugClient`, we can't get metadata of the regions on one tikv server because `get_all_meta_regions` is not implemented, and we will get better performance in this way.

wentaoxu:
after discussion with @zhangjinpeng1987 , I will change this to another way.

wentaoxu:
PTAL @BusyJay @zhangjinpeng1987

disksing:
closing for inactivity.

PR #226, opened 2018-01-30
support raft learner

Connor1996:


proto/pdpb.proto
```diff
@@ -240,6 +240,10 @@ message RegionHeartbeatRequest {
     uint64 keys_read = 9;
     // Approximate region size.
     uint64 approximate_size = 10;
+    // Pengding Learner peers are the non-voter peers that are catching up with leader's log
+    repeated metapb.Peer pending_learner_peers = 11;
```
hicqu: How about `pending_learners`?

proto/pdpb.proto
```diff
@@ -240,6 +240,10 @@ message RegionHeartbeatRequest {
...
+    // Pengding Learner peers are the non-voter peers that are waiting to be promoted to voter
+    repeated metapb.Peer complete_learner_peers = 12;
```
hicqu: Does we really need this? We can got these learners by `setdiff(region.learners, pending_learners)`.

hicqu: And I think you can also add `repeated Peer learners` in `metapb::Region` in this PR. @BusyJay what do you think?

hicqu:
rest LGTM. @BusyJay @siddontang PTAL thanks.

proto/pdpb.proto
Connor1996: get it.

hicqu:
LGTM.

hicqu:
ping @BusyJay

hicqu:
@Connor1996 Please re-geneerate code for resolving conflicts.

hicqu:
@overvenus @BusyJay PTAL thanks.

BusyJay:
LGTM, but I suggest not to merge it now. Let's develop base on this branch, and check what need to be tweaked.

hicqu:
According to latest talk, we can remove `ConfChangeType::PromoteLearnerNode`. Please improve that.

BusyJay:
I think we can merge this first, as no broken change is introduced.

PR #229, opened 2018-02-05
pdpb: add SplitRegion command.

disksing:
When tikv receives the SplitRegion command, it should split the region into 2 parts. It is encouraged to make the 2 region have the same size.


overvenus:
Would you mind add some descriptions?

disksing:
Here is the comment: https://github.com/pingcap/kvproto/pull/229/files#diff-302367292c5ac90dc53f4a17d1851bcbR286
@overvenus

BusyJay:
How to ensure the command won't be sent or executed multiple times concurrently?

disksing:
@BusyJay The command contains the region's epoch. tikv-sever may check if its version matches the command to decide whether to execute the command.

AndreMouche:
LGTM

PR #231, opened 2018-02-26
use eraftpb from raft-rs

BusyJay:
We now maintain the proto file in pingcap/raft-rs project, let's remove the one in this repository.


BusyJay:
PTAL

BusyJay:
ping

disksing:
LGTM.

PR #239, opened 2018-03-12
pd: add api to put/get kv on kvbase

wentaoxu:
as title says
PTAL @zhangjinpeng1987 @disksing @Connor1996


disksing:
LGTM.

disksing:
Now this PR has nothing to do with GC. I think the title need update.

wentaoxu:
ok @disksing

disksing:
`kvbase` is an internal concept of pd. I think better to use `etcd` instead.

PR #243, opened 2018-03-21
rawkv: support cf.

disksing:


disksing:
PTAL @zhangjinpeng1987

PR #249, opened 2018-04-13
add timestamp for heartbeat

nolouch:
use to report heartbeat interval time.


proto/pdpb.proto
```diff
@@ -354,6 +354,7 @@ message StoreHeartbeatRequest {
     RequestHeader header = 1;
 
     StoreStats stats = 2;
+    uint64 timestamp = 3;
```
disksing: `interval` or `duration`? Also please add some comments.

overvenus: There is a well-known `Timestamp`[1] message.

[1] https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp

disksing:
LGTM.

proto/pdpb.proto
```diff
@@ -242,8 +242,10 @@ message RegionHeartbeatRequest {
     uint64 keys_read = 9;
     // Approximate region size.
     uint64 approximate_size = 10;
-    // Unix timestamp when send the heartbeat.
-    uint64 timestamp = 11;
+    // The unix timestamp in seconds of the start of this period.
+    uint64 start_timestamp = 11;
```
overvenus: How about moving the `start_timestamp` and the `end_timestamp` into a message?

overvenus:
Cc https://github.com/pingcap/tikv/pull/2965

PR #251, opened 2018-04-18
Add a DebugMvccInfo to be compatible with old TiKV.

hicqu:
Compatibility between tikv-ctl and tikv is broken. So we can add `DebugMvccInfo` instead of use `MvccInfo` directly to be compatible with old tikv.


BusyJay:
So how can it be compatible with `MvccInfo`? Why is `MvccInfo` added in the first place? Is such compatibility really worthy?

PR #254, opened 2018-05-04
add split time field

Connor1996:
pd need to persist split time, and it is no need for tikv to fill this field.


proto/metapb.proto
```diff
@@ -51,6 +51,7 @@ message Region {
     bytes end_key = 3;
     RegionEpoch region_epoch = 4;
     repeated Peer peers = 5;
+    int64 last_split_timestamp = 6;
```
disksing: Add some comment.

Connor1996:
@huachaohuang If so, we can't guarantee not to merge new split region if pd leader change during importing. Is that acceptable?

huachaohuang:
@Connor1996 I think that's acceptable if you don't merge regions in `split-merge-interval` after leader change.

Connor1996:
@huachaohuang So persisting split time is still necessary, that is to say,  we can not just maintain the last split timestamp in memory.

huachaohuang:
@siddontang @BusyJay Do you think we should add this to the region metadata?

PR #255, opened 2018-05-04
Split importpb into import_kvpb and import_sstpb

huachaohuang:
We need to add some messages with the same name for ImportKV and
ImportSST, so we better split them now.


breeswish:
LGTM

PR #260, opened 2018-05-07
Add .gitattributes

### LINKED ISSUE #256: Add GitHub hint for generated files
https://github.com/github/linguist#generated-code

Reasons:

1. prevent from messing up our diff result (actually we only want to see .proto changes)
2. more accurate language statistics in the project homepage (currently we have Rust 99%)

breeswish:
- Automatically suppress diff for our generated files when reviewing PRs.
- Exclude generated files and vendor files from the GitHub's language detection.

Fix #256

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+/_vendor	linguist-vendored
+/pkg		linguist-generated
+/src		linguist-generated
+


breeswish:
@Hoverbear @BusyJay PTAL

breeswish:
It seems that the attribute is not taking effect... Maybe related to different directory format according to https://git-scm.com/docs/gitattributes   I'm going to open another PR, hope it works!

PR #262, opened 2018-05-08
Fix .gitattribute

breeswish:
Previous PR is https://github.com/pingcap/kvproto/pull/260  but it is not working.

According to [git-scm](https://git-scm.com/docs/gitattributes), we may need to replace directories with `foo/**` style. Hope it works.

@Hoverbear PTAL


BusyJay:
If it works, your fork should work too.

breeswish:
@busyjay Good idea! Let me fork it to have a check!

PR #264, opened 2018-05-11
Update eraftpb.proto

overvenus:
See more https://github.com/pingcap/raft-rs/pull/59

Close #265


BusyJay:
LGTM, CI fails due to the resolving policy of cargo, I think you can use `~1.4` to get it passed.

hicqu:
LGTM.

PR #265, opened 2018-05-14
Upgrade dependency raft version to 0.2

hicqu:
Upgrade dependency raft version from 0.1 to 0.2.


overvenus:
Missing codegen. Dup with #264.

PR #281, opened 2018-06-22
pdpb: support more methods for split region

hicqu:
Add a new region split method: at-logical-middle. This will split a region by logical middle, calculated with region start key and region end key.


hicqu:
@disksing @zhangjinpeng1987 PTAL thanks.

PR #283, opened 2018-06-28
add a new field `middle_key` for region properties.

hicqu:
Add a new field `middle_key` in region properties.


huachaohuang:
Why do you need to add this?

hicqu:
Sometimes we want to use `split_region` API. So we need to get approximate middle key by this RPC.

huachaohuang:
I remember there is a half split API?

hicqu:
TiKV only have a `split_region` API, which should be used with a given split_key. So how to calculate the key can be controled by caller. Maybe we can re-implement `HalfSplit` by the approximate key instead of scan region later, however let `get_region_properties` can also return the middle key is helpful for now.

hicqu:
ping @huachaohuang  @zhangjinpeng1987

proto/debugpb.proto
```diff
@@ -214,4 +214,6 @@ message GetRegionPropertiesRequest {
 
 message GetRegionPropertiesResponse {
     repeated Property props = 1;
-}
\ No newline at end of file
+    // The middle key of the region in encoded format, without ts padding.
+    bytes middle_key = 2;
```
zhangjinpeng1987: You can use 'middle_key_by_approximate_size' as an `props`'s name.

hicqu: Yes, so the PR can be closed now.

PR #284, opened 2018-07-01
debugpb: add bottommost level compaction

huachaohuang:
Make it more flexible for bottommost level compaction.


zhangjinpeng1987:
LGTM

PR #286, opened 2018-07-09
pdpb: use `approximate_write_keys` instead of  `approximate_rows`

AndreMouche:
Hi,
     This PR uses `approximate_write_keys` instead of  `approximate_rows` in `RegionHeartbeatRequest`

@disksing @overvenus PTAL


disksing:
why not `approximate_keys`? `approximate_write_keys` looks like the total written keys count in a period of time.

huachaohuang:
Yup, I think `approximate_keys` is enough, don't want to expose the `CF_WRITE` concept outside of TiKV.

huachaohuang:
LGTM

PR #287, opened 2018-07-11
pdpb: add check policy for split region

Connor1996:
Currently TiKV support get approximate middle key from properties, so when split region add a check-policy field to specify to use which way.


zhangjinpeng1987:
Please fix conflicts.

zhangjinpeng1987:
rest LGTM

zhangjinpeng1987:
LGTM

PR #288, opened 2018-07-11
Add a new field `store_id` in Debug::get_metrics RPC.

hicqu:
Add a new RPC to get store infomations from TiKV.


proto/debugpb.proto
```diff
@@ -224,4 +227,11 @@ message GetRegionPropertiesRequest {
...
     repeated Property props = 1;
-}
\ No newline at end of file
+}
+
+message StoreInfoRequest {
+}
+
+message StoreInfoResponse {
+    uint64 id = 1;
```
BusyJay: Do we need such standalone RPC? Can it be included in metrics response?

breeswish:
LGTM

BusyJay:
@LinuxGit PTAL

LinuxGit:
LGTM

PR #290, opened 2018-08-15
pd: add new error type for response

Connor1996:
Add a new error type `NOT_COMPATIABLE` for response, cause request may failed for version check.


proto/pdpb.proto
```diff
@@ -73,6 +73,7 @@ enum ErrorType {
     NOT_BOOTSTRAPPED = 2;
     STORE_TOMBSTONE = 3;
     ALREADY_BOOTSTRAPPED = 4;
+    NOT_COMPATIBLE = 5;
```
overvenus: How about `incompatible`?

For now, only `AskBatchSplit` will be refused, but there may be more incompatible requests, should we define an enum or message?

Connor1996:
PTAL again @overvenus  @nolouch  @disksing

nolouch:
LGTM

PR #291, opened 2018-08-17
Batch raft msg

hicqu:


hicqu:
Batch raft messages is not a good idea, because raft messages don't eat much gRPC threads. So it's not very effective. Batch messages between TiKV and clients is much better.

hicqu:
PTAL @zhangjinpeng1987 @overvenus

PR #295, opened 2018-09-21
*: add sync region service

nolouch:
add a `SyncRegions`  service.


proto/pdpb.proto
```diff
@@ -460,3 +462,18 @@ message UpdateGCSafePointResponse {
     uint64 new_safe_point = 2;
 }
 
+message SyncRegionRequest{
+    RequestHeader header = 1;
+    Member member = 2;
+    enum TP {
```
rleungx: better to add a comment to explain what it is?

proto/pdpb.proto
```diff
@@ -460,3 +462,18 @@ message UpdateGCSafePointResponse {
...
+	UNKNOW = 0;
```
rleungx: UNKNOWN

PR #299, opened 2018-10-09
vendor: switch to godep and update `gogo/protobuf`

nolouch:
update `gogo/protobuf` to `v1.1.1` and also ` grpc` to `1.12.0`


overvenus:
How do TiDB and PD build against this PR? Will it cause binary bloat?

Makefile
```diff
@@ -4,9 +4,6 @@ CURDIR := $(shell pwd)
 
 KEEP_FILE := '**/*.proto,**/*.sh'
```
overvenus: It can be removed.

Gopkg.toml
```diff
@@ -0,0 +1,25 @@
...
+[[constraint]]
+  name = "github.com/golang/protobuf"
+  source = "https://github.com/golang/protobuf.git"
+  version = "1.2.0"
+
+[[constraint]]
+  branch = "master"
+  name = "golang.org/x/net"
+
+[prune]
```
overvenus: How about pruning `non-go = true`?

Makefile
```diff
@@ -4,9 +4,6 @@ CURDIR := $(shell pwd)
...
 
-export PATH := $(CURDIR)/_vendor/bin:$(PATH)
-export GOPATH := $(CURDIR)/_vendor
```
overvenus: By removing these lines, generate_go may install few go tools into your real gopath and use wrong protobuf and grpc compiler.

Gopkg.toml
nolouch: some licenses should be kept.

zhexuany:
dep part looks good to me.

nolouch:
@overvenus I test `tidb` with this change and no binary bloat.

nolouch:
close by: https://github.com/pingcap/kvproto/pull/300 
new pr will trigger the ci.

PR #300, opened 2018-10-15
vendor: switch to godep and update gogo/protobuf

nolouch:
update `gogo/protobuf` to `v1.1.1` and also ` grpc` to `1.12.2`.


Makefile
```diff
@@ -2,35 +2,22 @@
...
 	# Standalone GOPATH
 	./generate_go.sh
 	go build ./pkg/...
 
-rust:
+rust: init
 	./generate_rust.sh
 	cargo check
 
 update_go_pkg:
```
rleungx: better to use a hyphen `-` instead of an underscore `_`?
https://style-guides.readthedocs.io/en/latest/makefile.html

PR #302, opened 2018-10-23
generate cpp code used for ch raft log sync

hanfei1991:


overvenus:
Could you test(build) the cpp target on CI?

PR #304, opened 2018-10-24
add reverse option in RawBatchScanRequest

kamijin-fanta:
- support backward scan
- for https://github.com/tikv/tikv/issues/3705


MyonKeminta:
In fact, I suggest you also update RawScan because currently it seems that the go client doesn't support RawBatchScan...

kamijin-fanta:
also possible to change both.
raw_batch_scan internally use raw_scan.

MyonKeminta:
Yes, I suggest that you add the `reverse` field in both `RawScanRequest` and `RawBatchScanRequest`

kamijin-fanta:
oh sorry. change it like that.

kamijin-fanta:
changed it

kamijin-fanta:
@overvenus @MyonKeminta changed it. please review.

PR #306, opened 2018-10-28
Add end_key to kv scan request

MyonKeminta:


proto/kvrpcpb.proto
```diff
@@ -98,6 +98,10 @@ message ScanRequest {
     uint64 version = 4;
     bool key_only = 5;
     bool reverse = 6;
+    // For compatibility, when scanning forward, `start_key` is the inclusive lower_bound and `end_key`
+    // is the exclusive upper_bound; when scanning backward, `start_key` is the exclusive upper_bound
+    // and `end_key` is the inclusive lower_bound.
+    bytes end_key = 7;
```
siddontang: maybe we can use `[start_key, end_key)` or `[end_key, start_key)` for simple explanation. 

But, for backward, maybe start_key is inclusive?

MyonKeminta: In the current implementation in TiKV, it's exclusive.

MyonKeminta:
@siddontang PTAL again @zhangjinpeng1987 PTAL

proto/kvrpcpb.proto
zhangjinpeng1987: Seems we don't use reverse scan?

MyonKeminta:
@siddontang PING

PR #311, opened 2018-11-06
pdpb: region syncer add index filed

nolouch:
add the index to sync the records.


proto/pdpb.proto
```diff
@@ -465,10 +465,12 @@ message UpdateGCSafePointResponse {
 message SyncRegionRequest{
     RequestHeader header = 1;
     Member member = 2;
+    uint64 start_index = 3;
```
disksing: Add some comment to describe what does it used for?

disksing:
LGTM.

PR #317, opened 2018-11-27
use retool and modules to manage dependency

disksing:


nolouch:
Can the `generate_rust.sh` work?

disksing:
@nolouch Of cause, the CI will be broken if not.

overvenus:
Drop vendor may cause unexpected error, e.g., someone force push to master `golang.org/x/text/`.

PR #333, opened 2018-12-05
errorpb: add details to StoreNotMatch

BusyJay:
It helps to diagnose implementation problems.


BusyJay:
After discussing with @disksing, we decided that leave it to future when it's really necessary.

BusyJay:
To reduce redundant information, it's better to add it into struct explicitly.

disksing:
LGTM.

PR #343, opened 2019-01-21
Push down precondition to TiKV

zhangjinpeng1987:
Push down precondition to TiKV, so for INSERT statement with unique constraint, the batch_get is not necessary any more.


zhangjinpeng1987:
@disksing @MyonKeminta PTAL

proto/kvrpcpb.proto
```diff
@@ -18,11 +18,16 @@ message LockInfo {
...
+message AlreadyExist {
+    bytes key = 1;
+}
+
 message KeyError {
     LockInfo locked = 1; // Client should backoff or cleanup the lock then retry.
     string retryable = 2; // Client may restart the txn. e.g write conflict.
     string abort = 3; // Client should abort the txn.
     WriteConflict conflict = 4; // Write conflict is moved from retryable to here.
+    AlreadyExist already_exist = 5; // Key already exists
```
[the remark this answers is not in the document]
zhangjinpeng1987: Currently we only use should_not_exist.

proto/kvrpcpb.proto
```diff
@@ -18,11 +18,23 @@ message LockInfo {
     uint64 lock_ttl = 4;
 }
 
+message AlreadyExist {
+    bytes key = 1;
+}
+
+message NotEqualTo {
```
MyonKeminta: How about just `NotEqual` or `NotMatch`?

proto/kvrpcpb.proto
```diff
@@ -18,11 +18,23 @@ message LockInfo {
...
+    bytes key = 1;
+    bytes value = 2;
+    bytes should_equal_to = 3;
```
MyonKeminta: How about something like `expected_value`?

PR #345, opened 2019-01-25
kvrpcpb: add contract proto for data/index consistency

tiancaiamao:


proto/kvrpcpb.proto
```diff
@@ -33,8 +37,7 @@ message KeyError {
     string retryable = 2; // Client may restart the txn. e.g write conflict.
     string abort = 3; // Client should abort the txn.
     WriteConflict conflict = 4; // Write conflict is moved from retryable to here.
-    AlreadyExist already_exist = 5; // Key already exists
-    NotEqualTo not_equal_to = 6; // Value not equal to specified value
+    PreconditionError precondition = 5; // PreconditionResponse is not an error from MVCC's perspective.
```
zhangjinpeng1987: s/PreconditionResponse/PreconditionError

zhangjinpeng1987: s/precondition/precondition_err

tiancaiamao:
Now this one updated and it has nothing to do with https://github.com/pingcap/kvproto/pull/353
PTAL @coocood @disksing

coocood:
LGTM

proto/kvrpcpb.proto
```diff
@@ -133,19 +131,18 @@ enum Op {
...
+     MustNotExist = 2;
 }
 
 message Mutation {
     Op op = 1;
     bytes key = 2;
     bytes value = 3;
-    // Write only when the precondition satisfied
-    Precondition precondition = 4;
+    // Write only when the assertion holds
```
coocood: We write even if assertion failed, only log the error and metrics.

proto/kvrpcpb.proto
```diff
@@ -133,19 +131,18 @@ enum Op {
...
 
-message Precondition {
-    // Should not exist
-    bool should_not_exist = 1;
-    // Value must equal to
-    bytes equal_to = 2;
+enum Assertion {
+     NoCheck = 0;
+     MustExist = 1;
+     MustNotExist = 2;
```
coocood: Remove the `Must` is better.
And I think `None` is better than `NoCheck`.

coocood:
LGTM

tiancaiamao:
PTAL @disksing

PR #353, opened 2019-02-01
add Insert operation

zhangjinpeng1987:
Signed-off-by: zhangjinpeng1987 <zhangjinpeng@pingcap.com>

To keep Put operation pure, add `Insert` operation. `Insert` operation has a constraint that key should not exist before, or will return `AlreadyExist` error.


zhangjinpeng1987:
@disksing @coocood @tiancaiamao PTAL

coocood:
LGTM

zhangjinpeng1987:
@disksing PTAL
@coocood PTAL again

PR #373, opened 2019-04-01
Add notify_only field to kv delete_range request and raft delete_range request

MyonKeminta:

When `notify_only` is set, tikv will not actually performing deleting on apply but will wait for an unsafe_destroy_range request. It's used to notify that an unsafe_destroy_range is going to be executed.


overvenus:
Why not change tikv's behavior, do nothing when it applies delete range command so that we do not need to add a new command.

MyonKeminta:
@overvenus I thought it's not a good idea to beak `delete_range`'s behavior

MyonKeminta:
@overvenus PTAL again.

MyonKeminta:
@overvenus @zhangjinpeng1987 PTAL, thanks!

MyonKeminta:
@overvenus @zhangjinpeng1987 PTAL thanks!
Also please check #408

nrc:
Note for reviewers: we shouldn't merge PRs like this which drop all the formatting of the generated files. It makes for huge diffs and difficult to follow `git blame`. If you have Rustfmt installed, then the generated code will be automatically formatted. To install: `rustup component add rustfmt`.

Unfortunately this is not easy to check in CI since `rustfmt --check` will not check the `#include`ed files, but I'll try and figure something out.

BusyJay:
Maybe we should just put all the generated code in $OUT directory instead of src.

PR #377, opened 2019-04-13
make debug format pretty again

BusyJay:
Bring back the pretty print of struct, raw debug format seems annoy everyone.

### DIFF
--- a/src/protobuf/raft_serverpb.rs
+++ b/src/protobuf/raft_serverpb.rs
@@ -21,7 +21,7 @@
 use protobuf::Message as Message_imported_for_functions;
 use protobuf::ProtobufEnum as ProtobufEnum_imported_for_functions;
 
-#[derive(PartialEq,Clone,Default,Debug)]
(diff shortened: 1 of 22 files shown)


src/protobuf/raft_serverpb.rs
```diff
@@ -507,13 +507,19 @@ impl ::protobuf::Clear for RaftMessage {
     }
 }
 
+impl ::std::fmt::Debug for RaftMessage {
```
overvenus: Which controls the codegen behavior? Can we ensure everyone generates the right `Debug` implementation?

BusyJay: I remove the check in codegen logic. https://github.com/BusyJay/rust-protobuf/commit/7da1412e210cfc50e881159428c4cf708653d2ed#diff-6499709ed714601b45000c51bfa00217

BusyJay:
Turns out that pretty format requires reflection, I need to introduce the format in other way.

PR #380, opened 2019-04-20
add watch rpc

bradyjoestar:
https://github.com/pingcap/pd/issues/1478


nolouch:
looks good. you can use `replace` in go mod firstly in your code, and we can adjust the proto later.

bradyjoestar:
Sure, it's better to merge it when watch service in `pd` is stable. :smile:

PR #382, opened 2019-04-25
supports pessimistic transaction

coocood:


proto/deadlock.proto
```diff
@@ -0,0 +1,51 @@
...
+	repeated WaitForEntry entries = 1 [(gogoproto.nullable) = false];
+}
+
+message WaitForEntry {
+	// The transaction id that is waiting.
+	uint64 txn = 1;
+	// The transaction id that is being waited for.
+	uint64 wait_for_txn = 2;
+	// The hash value of the key is being waited for.
+	uint64 key_hash = 3;
```
zhangjinpeng1987: If there is hash conlfict, does it matter?

proto/deadlock.proto
```diff
@@ -0,0 +1,51 @@
...
+}
+
+enum DeadlockRequestType {
+	Detect = 0;
+	// CleanUpWaitFor cleans a single entry the transaction is waiting.
+	CleanUpWaitFor = 1;
+	// CleanUp cleans all entries the transaction is waiting.
+	CleanUp = 2;
```
zhangjinpeng1987: Maybe CleanUpAll?

proto/deadlock.proto
coocood: Then it will timeout, it's very rare, so I think it's acceptable.

proto/kvrpcpb.proto
```diff
@@ -151,13 +152,32 @@ message PrewriteRequest {
     uint64 start_version = 4;
     uint64 lock_ttl = 5;
     bool skip_constraint_check = 6;
+    // For pessimistic transaction, some mutations don't need to be locked, for example, non-unique index key.
+    repeated bool is_pessimistic_lock = 7;
```
zhangjinpeng1987: Embeded in Mutation is better?

coocood: That would increase the unmarshal overhead for the non-pessimistic transactions.

proto/deadlock.proto
coocood: `CleanUpAll` can be misunderstood as clean up all transactions in the detector.

proto/deadlock.proto
coocood: No, in this case, it doesn't matter, because there is definitely a deadlock.

coocood: The key_hash can be used to make deadlock only retry a single statement.

youjiali1995:
LGTM

proto/kvrpcpb.proto
zhangjinpeng1987: Why?

coocood: the Mutation struct is bigger.

tiancaiamao:
LGTM

PR #384, opened 2019-04-26
Add messages used to test batching

benpigchu:


ice1000:
Could you please fix the conflicts?

benpigchu:
@ice1000 not now. Keep this open at this moment.

proto/tikvpb.proto
```diff
@@ -125,15 +127,29 @@ message BatchCommandsResponse {
...
+message BatchCommandTestRequest {
+    // Id of the test request
+    uint64 test_id = 1;
+    // The time between receiving the request and sending the response
+    uint64 delay_time = 2;
+}
+
+message BatchCommandTestResponse {
+    // Id of the test response, should be same with the one in the request
+    uint64 test_id = 1;
```
hicqu: The response is generated out of gRPC threads. So you can add
* `reply_time` which is for when the response is generated
* `reply_send_time` which is for when the response is sent out

hicqu:
rest LGTM. After we get 2 LGTM, I think we can push a `3.0` branch to pingcap/kvproto.

ice1000:
image
FYI You can actually change the base branch.

hicqu:
LGTM. Now we can consider how to use it.

PR #385, opened 2019-04-28
[DNM] add txn-size in lock & resolve specified keys

zhangjinpeng1987:
Signed-off-by: zhangjinpeng1987 <zhangjinpeng@pingcap.com>

- ResolveLock support only resolve specified keys.
- Add txn-size in lock, so we can choose how to resolve the keys when read request encounter locks.


zhangjinpeng1987:
Use https://github.com/pingcap/kvproto/pull/386.

PR #396, opened 2019-05-14
pdpb: add ScanRegions

disksing:


proto/pdpb.proto
```diff
@@ -204,6 +206,17 @@ message GetRegionByIDRequest {
 
 // Use GetRegionResponse as the response of GetRegionByIDRequest.
 
+message ScanRegionsRequest {
+    RequestHeader header = 1;
+    bytes region_key = 2;
```
MyonKeminta: It seems that there should be an empty line above here?

disksing: ok

proto/pdpb.proto
```diff
@@ -204,6 +206,19 @@ message GetRegionByIDRequest {
...
+    RequestHeader header = 1;
+
+    bytes region_key = 2;
+    int32 limit = 3;
+}
+
+message ScanRegionsResponse {
+    ResponseHeader header = 1;
+    repeated metapb.Region regions = 2;
+    repeated metapb.Peer leaders = 3;
```
zhangjinpeng1987: How about region has no leader?

proto/pdpb.proto
zhangjinpeng1987: I think `start_key` is a better name.

disksing: ok

proto/pdpb.proto
disksing: It will add empty value to make regions and leaders are matched.

proto/pdpb.proto
```diff
@@ -204,6 +206,19 @@ message GetRegionByIDRequest {
...
+message ScanRegionsRequest {
+    RequestHeader header = 1;
+
+    bytes start_key = 2;
+    int32 limit = 3;
+}
+
+message ScanRegionsResponse {
+    ResponseHeader header = 1;
+    repeated metapb.Region regions = 2;
```
MyonKeminta: Oh, here should be an empty line too

zhangjinpeng1987:
@disksing Please cherry-pick to release-3.0 branch, thx.

PR #397, opened 2019-05-16
 Reference the Prost branch of raft-rs (and formatting)

nrc:
This PR changes the git rev to point at the Head of the prost branch of Raft, rather than at 0.4.1. It also adds formatting of the Protobuf generated files.

PTAL @overvenus


Cargo.toml
```diff
@@ -19,8 +19,8 @@ prost-derive = "0.5"
 bytes = "0.4"
 futures = "0.1"
 grpcio = { version = "0.5.0-alpha.1", features = ["prost-codec"] }
-raft = { git = "https://github.com/pingcap/raft-rs.git", rev = "7c5bdfc58d23860ad6111fbbfe1d4db1d4c3d894" }
+raft = { git = "https://github.com/pingcap/raft-rs.git", rev = "8e953b0d0c9cd2f98c94a42edc48ae3f09e967ff" }
```
overvenus: After this PR is merged, TiKV must use `8e953b0d0c9cd2f98c94a42edc48ae3f09e967ff` too, right?

nrc: No, TiKV is using a branch of KvProto (and is not using Prost yet, so still needs the non-Prost Raft)

nrc:
@overvenus @zhangjinpeng1987 PTAL

PR #404, opened 2019-05-27
add Deadlock Key error and for_update_ts in Prewrite

coocood:
- `Deadlock` key error will be used for pessimistic transaction single statement rollback.
- `for_update_ts` in `PrewriteRequest`  will be used to check if a conflict lock is already committed.


coocood:
@zhangjinpeng1987 @youjiali1995 @tiancaiamao PTAL

youjiali1995:
Please cherry pick it to branch 'release-3.0' after merged.

PR #407, opened 2019-06-03
Enable Refresh Lock Request

birdstorm:
Provide `RefreshLockRequest` for long-lasting transactions. Users can now use a lock with short TTL and refresh it at ease.

Merge `release-3.0` branch to `master` so that we will be using `master` branch of `kvproto` in `TiKV` instead of `release-3.0`.


proto/tikvpb.proto
```diff
@@ -70,7 +71,7 @@ service Tikv {
 message BatchCommandsRequest {
```
MyonKeminta: I think it's better to add it to `BatchCommands`, so that it can be batched with other requests to save RPC.

proto/kvrpcpb.proto
```diff
@@ -295,6 +295,20 @@ message ResolveLockResponse {
     KeyError error = 2;
 }
 
+message RefreshLockRequest {
+    Context context = 1;
+    uint64 start_version = 2;
+    bytes key = 3;
+    uint64 ttl = 4;
+    uint64 txn_size = 5;
```
youjiali1995: What the use of txn_size?

birdstorm:
@MyonKeminta @youjiali1995 I have moved this PR to #414 so that it will base on release-3.0. Comments are now addressed in #414.

This PR will be closed.

I must say that it is rather strange that TiKV’s `master` now bases on kvproto’s `release-3.0` branch.

MyonKeminta:
I think this PR should continue. We will switch TiKV's kvproto back to master in the future.

PR #410, opened 2019-06-17
*: report term in region heartbeat

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>
report the raft term about the region in pd heartbeat.


breeswish:
Please apply change to master as well.

Issue #280, opened 2018-06-21
Publishing?


Hoverbear:
I'm interested in publishing this crate so that it can be used as part of the upcoming TiKV client in Rust (which will be published on crates.io).

This already builds cleanly on nightly and has no illegal dependencies, so there should be nothing blocking publishing this.

Hoverbear:
@siddontang PTAL

Hoverbear:
We need to seriously consider this with the `tikv-client` for Rust. We cannot publish the client without all of its dependencies also being published.

nrc:
Some things we need to do/think about:

- [ ] Use a crates.io version of Raft (I think this just needs us to finish the Prost migration and other Raft versioning issues.)
- [ ] Consider the name - 'kvproto' is not very self-explanatory (tikv_proto is suggested above, I prefer tikv-proto (consistent with rust-client, which should probably be tikv-client.rs or something)
- [ ] Consider and document the process for releases
- [ ] What should a Rust crate look like? We probably should not include the Go and CPP code in the distributed tarball, for example.
- [ ] building - are we happy with the current process and requirements for building (see also #354)
- [ ] think about the project org maybe? (Some things worth considering: why is Raft the source of truth for it's Protos, but this project is for the other protos? Why would a proto be here vs elsewhere? Should we be committing the generated code? Is having a multi-language project the right approach (especially if we add more languages in the future)?

Probably more...

Hoverbear:
@nrc Do you think extracting only the tikv related protos and making them a dependency here would be a good first step to enable this? Maybe a `tikv-api` crate? Then we could publish that.

nrc:
> Do you think extracting only the tikv related protos and making them a dependency here would be a good first step to enable this? Maybe a tikv-api crate? 

I think that is probably a good idea. I wonder whether the Raft protos should be part of that?

ice1000:
These protos were too big for Raft (which is relatively small) to depend on, but the situation may change after prost

BusyJay:
No, raft is a standard library that should keep its own protocols.

nrc:
I don't understand either of those points, sorry.

> These protos were too big for Raft (which is relatively small) to depend on

Why is size an issue?

> No, raft is a standard library that should keep its own protocols.

What do you mean by "standard" (and by implication why are TikV, etc., not standard?)

BusyJay:
Sorry, it should be standalone. :rofl:

ice1000:
[...]

PR #411, opened 2019-06-18
Depend on raft-proto crate instead of raft

nrc:
Means we don't need a Git dependency and can publish KvProto on crates.io (#280).

Also re-generates the Prost code to avoid hundreds of warnings.

PTAL @Hoverbear @BusyJay


Cargo.toml
```diff
@@ -19,7 +19,7 @@ prost-derive = "0.5"
 bytes = "0.4"
 futures = "0.1"
 grpcio = { version = "0.5.0-alpha.1", features = ["prost-codec"] }
-raft = { git = "https://github.com/pingcap/raft-rs.git", rev = "8e953b0d0c9cd2f98c94a42edc48ae3f09e967ff" }
+raft-proto = "0.1"
```
Hoverbear: Is this the right revision for TiKV? https://github.com/tikv/tikv/blob/master/Cargo.toml#L109

nrc: TiKV is using a branch of KvProto at the moment since master is Prost-only, but this does break my Prost branch. I'll address it...

Cargo.toml
```diff
@@ -19,7 +19,7 @@ prost-derive = "0.5"
 bytes = "0.4"
 futures = "0.1"
 grpcio = { version = "0.5.0-alpha.1", features = ["prost-codec"] }
-raft = { git = "https://github.com/pingcap/raft-rs.git", rev = "8e953b0d0c9cd2f98c94a42edc48ae3f09e967ff" }
+raft-proto = "0.4"
```
Hoverbear: Can you explain the versioning strategy you expect?

ice1000: By following raft-rs?

nrc: Yes, matching with Raft. (Although if we do something different, it doesn't matter much).

nrc:
> I really wish we can put the generated code inside OUT directory so it doesn't have to be included in version control. No one takes much care about it.

I'd be happy to have that setup. Do you know why we build into src at the moment?

Hoverbear:
@nrc Please, go ahead :)

PR #413, opened 2019-06-20
Fix formatting of generated Rust code

MyonKeminta:


MyonKeminta:
I think this can be fixed by further PRs to muster. This PR is not needed.

PR #414, opened 2019-06-20
support refresh lock request

birdstorm:
Support refresh lock request.
Previous PR is #407 . This PR changes base to release-3.0 because two branches diverged.


proto/tikvpb.proto
```diff
@@ -83,22 +84,23 @@ message BatchCommandsRequest {
...
-            kvrpcpb.RawDeleteRangeRequest RawDeleteRange = 20;
-            kvrpcpb.RawBatchScanRequest RawBatchScan = 21;
-
-            coprocessor.Request Coprocessor = 22;
-            kvrpcpb.PessimisticLockRequest PessimisticLock = 23;
-            kvrpcpb.PessimisticRollbackRequest PessimisticRollback = 24;
+            kvrpcpb.RefreshLockRequest RefreshLock = 11;
+            kvrpcpb.GCRequest GC = 12;
+            kvrpcpb.DeleteRangeRequest DeleteRange = 13;
+
```
MyonKeminta: These IDs are changed. Won't this cause incompatibility?

proto/kvrpcpb.proto
```diff
@@ -142,9 +142,9 @@ enum Op {
 }
 
 enum Assertion {
-     None = 0;
-     Exist = 1;
-     NotExist = 2;
+    None = 0;
```
overvenus: It's better to define `UNKNOWN = 0`, https://developers.google.com/protocol-buffers/docs/proto3#enum

proto/tikvpb.proto
birdstorm: ok, I will append this request to the end of batch commands and maintain the original order.

birdstorm:
> It seems that generated prost code should not be in release-3.0 branch? @overvenus PTAL

Should I delete the generated files?

proto/kvrpcpb.proto
birdstorm: ok

youjiali1995:
I think you can file a new PR based on release-3.0.😂

overvenus:
> Should I delete the generated files?

Just `make rust go`, files will be updated automatically.

birdstorm:
> > Should I delete the generated files?
> 
> Just `make rust go`, files will be updated automatically.

I think I have already done that.

> I think you can file a new PR based on release-3.0.😂

This PR already bases on release-3.0.

overvenus:
Maybe you forget to push commits? Comments are not yet addressed.

proto/kvrpcpb.proto
birdstorm: @overvenus hey, do i need to modify its name in this PR?

overvenus: Yes, better to follow the style, also please address the [comment](https://github.com/pingcap/kvproto/pull/414/files#r300267740), change tag ID is not allowed since it's a breaking change.

birdstorm:
@overvenus PTAL

youjiali1995:
How did you generate prost code? It won't be generated in release-3.0.

> > It seems that generated prost code should not be in release-3.0 branch? @overvenus PTAL
> 
> Should I delete the generated files?

Yes.

birdstorm:
As commented in https://github.com/tikv/tikv/pull/4831, refreshing lock is incompatible with TiDB’s current resolve lock logic. This PR is closed.

PR #415, opened 2019-07-02
add an empty request in batch commands.

hicqu:
Add a new request `Empty` in `BatchCommands` API.
The new request is generally used for tests: clients can send such requests to TiKVs, and TiKVs won't do any special things for them but only send back such responses. So we can focus on the transport layer.


breeswish:
Hi, please update your change to master.

nrc:
I ported to master in https://github.com/pingcap/kvproto/pull/418

PR #416, opened 2019-07-02
Relax the required protobuf version

nrc:
cc https://github.com/tikv/tikv/issues/4864

PTAL @AndreMouche @siddontang

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ prost-codec = ["prost", "prost-derive", "bytes"]
 regenerate = []
 
 [dependencies]
-protobuf = { version = "~2.0", optional = true }
+protobuf = { version = "2", optional = true }
 prost = { version = "0.4", optional = true }
 prost-derive = { version = "0.4", optional = true }
 bytes = { version = "0.4.11", optional = true }


BusyJay:
This PR makes the replace not valid anymore.

https://github.com/pingcap/kvproto/blob/release-3.0/Cargo.toml#L32-L34

ice1000:
https://github.com/pingcap/kvproto/blob/d9830856b531614da3b6924390218371cb024a15/Cargo.toml#L32-L34

ice1000:
What do you mean by "not valid"? Since CI passes I don't quite understand you here.

BusyJay:
Changing the version from "~2.0" to "2" will make kvproto depends on rust-protobuf 2.7, which can't be replaced by the configuration anymore. If you try to build the project locally, you should see the warnings at the very beginning.

BusyJay:
Actually, those warning shows up in CI as well. 🤣

ice1000:
Does changing the replaced version from `2.0.6` to `2.7` sound like a plan?

BusyJay:
The patch was updated according to 2.0.6, they need to be refactored to 2.7 first.

ice1000:
Our previously specified version (which is not `2.0.6` that is hard-coded here, but `~2.0` which may change by time) is causing a versioning conflict in tikv so I'd say we should either update the patch or lock the version to 2.0.6 everywhere (tipb, tikv, kvproto, grpc-rs, raft-rs, etc.). What do you think? @nrc @BusyJay

nrc:
:-( We should not be using replace/patch in production - they are extremely fragile and intended for use in development or for very temporary fixes.

Note that even without changing the protobuf version the `replace` would break if a `2.0.7` was released.

We should either upstream the changes or properly fork rust-protobuf. If the latter then we should accept and plan for the maintenance burden. In any case we should be using a git dep (or better publish the forked crate), rather than using `replace`.

BusyJay:
> should not be using replace/patch in production

It's OK to use replace in production. The problem here is that replace is used for a library project instead of a binary project. The former doesn't have a lock file to lock all the version of all dependencies.

Here uses replace because the patch is necessary only when generating the code. Switch to fork version will be too complicated as all the dependencies need to be changed especially raft-proto.

I will prefer to keep the replace way and update the patch to version 2.7. And maybe figure out a way to let cargo report error on unused replace.

In the long term, if we switch to prost, then there will be no replace anymore; if we keep using rust-protobuf, better extract the patch in a plugable way. Or rollback to the old days when we still used binary to generate codes. In that case, a forked version can be used without changing any dependency.

PR #417, opened 2019-07-04
regenerating codes and update replace versions

BusyJay:
This PR fixes two problems introduced by #416, including stale source code and invalid replace configurations.

It also adds two more CI checks to prevent similar problems from happening again.


BusyJay:
It's marked as DNM as I'm verifying revert regenerating code will make CI fail.

BusyJay:
It works as expected now.

PTAL @nrc @AndreMouche @overvenus

breeswish:
What should we do after switching TiKV to this branch?

```
error[E0425]: cannot find value `VERSION_2_7_0` in module `protobuf`
  --> /Users/breezewish/.cargo/git/checkouts/kvproto-836cd5e20b5f8503/ed13754/src/protobuf/coprocessor.rs:27:49
   |
27 | const _PROTOBUF_VERSION_CHECK: () = ::protobuf::VERSION_2_7_0;
   |                                                 ^^^^^^^^^^^^^ not found in `protobuf`
```

BusyJay:
TiKV should update rust-protobuf to 2.7.0.

Makefile
```diff
@@ -14,10 +14,10 @@ go: init
 	GO111MODULE=on go build ./pkg/...
 
 rust: init
-	cargo check --features regenerate
+	@cargo check --features regenerate 2>&1 | tee /dev/fd/2 | grep -qvz warning || (echo Please fix warnings; exit 1)
```
nrc: Could use `RUSTFLAGS="-Dwarnings" cargo check`, rather than grep here.

.circleci/config.yml
```diff
@@ -37,11 +37,15 @@ jobs:
...
             unzip protoc-$PROTOC_VERSION-linux-x86_64.zip -d $HOME/.local
+            rm protoc-$PROTOC_VERSION-linux-x86_64.zip
             sudo apt install -y cmake golang
       - run:
           name: Test Rust
           command: make rust
           no_output_timeout: 1800s
+      - run:
+          name: Check if all files are regenerated
+          command: git update-index --refresh &>/dev/null || (git diff; echo please "make rust" before submitting a PR; exit 1)
```
nrc: IIRC, `file_descriptor_proto_data` is non-deterministic and can change when built on different platforms (I'm not sure exactly when it changes, but I have seen it change often when rebuilding the protos). Seems like that would trigger this check?

BusyJay: I don't think such field exist anymore after lite-runtime is enabled.

Makefile
BusyJay: This is exactly what I have tried at the beginning. But turns out that the warning is output by cargo shell rather than rust, so the flag will not work.

nrc: Ah, of course. Could you add a comment please so we remember that in the future?

PR #418, opened 2019-07-08
Update kvproto with Prost changes and port #415

nrc:
Use a Git dep for Prost so we get some optimisations which are not present in the current release.

PTAL @overvenus @Hoverbear


nrc:
CI will fail until https://github.com/pingcap/raft-rs/pull/258 lands and we publish a new version of raft-proto

nrc:
PTAL @Hoverbear @overvenus @hicqu (for the third commit)

PR #426, opened 2019-07-16
Support both Prost and rust-protobuf

nrc:
Should probably block on https://github.com/pingcap/raft-rs/pull/262 landing to avoid the git dep to my branch. This will allow us to work on both Prost and rust-protobuf on master.


ice1000:
How are we supposed to review this?

nrc:
> How are we supposed to review this?

You can ignore the code in src/protobuf - it is all auto-generated. If you are very keen you can pull the branch, delete those files run `cargo build --features gen` and check they are the same.

The rest of the PR is not too huge.

build.rs
```diff
@@ -68,6 +69,18 @@ fn main() {
...
+    generate_protobuf_files(&file_names, "src/protobuf");
+    let mod_names = module_names_for_dir("src/protobuf");
+
+    let out_file_names: Vec<_> = mod_names
+        .iter()
+        .map(|m| format!("src/protobuf/{}.rs", m))
+        .collect();
+    let out_file_names: Vec<_> = out_file_names.iter().map(|f| &**f).collect();
+    replace_read_unknown_fields(&out_file_names);
+    generate_lib_file(&mod_names);
```
ice1000: and then ` protobuf_build::rustfmt(Path::new("src/protobuf.rs"));`?

BusyJay:
Close as it's duplicated with #427.

PR #427, opened 2019-07-18
*: support rust-protobuf by default

BusyJay:
And support prost optionally. prost build is broken for now and should
be fixed later.

This is part of effort to make master of TiKV be compatible with tipb, kvproto and raft-rs.


breeswish:
Does it mean that generated protobuf files are no longer carried in the repository since this PR?

BusyJay:
Yes, it's generated at compile time.

nrc:
CI seems to be failing because of a bad Git ref for protobuf-build. Why is Prost broken? Otherwise it looks fine to me, but I think we should resolve the question of generated code before merging.

BusyJay:
Now, prost build is OK.

BusyJay:
Although CI is passed, but I suggest not to merge it for now. Better fix the problems found in pingcap/grpc-rs#346 first, otherwise TiKV will probably meet the similar error.

BusyJay:
OK, I think it's OK to merge it now. We can fix the problem before making tikv depend on master of kvproto.

PR #428, opened 2019-07-19
cherry pick proto changes from release-3.0

BusyJay:
So that master of TiKV can depend on the master of kvproto.


shenli:
Could you link the original PR here?

BusyJay:
pingcap/kvproto#410 and pingcap/kvproto#424

PR #436, opened 2019-08-07
import_kvpb: get more information from rpc

lance6716:
This PR is to solve https://github.com/tikv/importer/issues/4 .


lance6716:
PTAL @kennytm @BusyJay

lance6716:
@overvenus PTAL

proto/import_kvpb.proto
```diff
@@ -46,6 +46,10 @@ service ImportKV {
     rpc CleanupEngine(CleanupEngineRequest) returns (CleanupEngineResponse) {}
     // Compact the target cluster for better performance.
     rpc CompactCluster(CompactClusterRequest) returns (CompactClusterResponse) {}
+    // Get current version and commit number
+    rpc GetVersion(GetVersionRequest) returns (GetVersionResponse) {}
+    // Get importer metrics
+    rpc GetMetrics(GetMetricsRequest) returns (GetMetricsResponse) {}
```
overvenus: Dup with `http://pd_or_tikv_addr:port/metrics`?

kennytm: We haven't exposed the StatusServer yet since it depends on TiKvConfig (and IIRC it was in a private module previously; it is now public). 

Even if we do expose the StatusServer there still need to be a way to tell Lightning the `status_port` (which is different from `port`).

lance6716:
Ping @kennytm @overvenus 
This repo needs 2 approving reviews, LGT1 doesn't count here 😛

kennytm:
Ping @overvenus. Is the current change acceptable? Or do you suggest we must go through the HTTP interface?

overvenus:
Sorry for the late reply, it just feels strange though I don't have a strong objection.

We still need to write a parser if `GetMetrics` output raw Prometheus metrics, right?
Also, where do we save these metrics?

lance6716:
> Sorry for the late reply, it just feels strange though I don't have a strong objection.
> 
> We still need to write a parser if `GetMetrics` output raw Prometheus metrics, right?
> Also, where do we save these metrics?

Lightning will visualize these metrics on its log and Web interface.

Seems that `http://pd_or_tikv_addr:port/metrics` will also generate raw metrics which still need parser, so expose HTTP API will require almost same work.

proto/import_kvpb.proto
```diff
@@ -46,6 +46,10 @@ service ImportKV {
     rpc CleanupEngine(CleanupEngineRequest) returns (CleanupEngineResponse) {}
     // Compact the target cluster for better performance.
     rpc CompactCluster(CompactClusterRequest) returns (CompactClusterResponse) {}
+    // Get current version and commit number
```
overvenus:
```suggestion
    // Get current version and commit hash
```

PR #437, opened 2019-08-08
Add `not_found` field to distinguish a non-existent key from zero-length value

sticnarf:
Currently, when we get or batch get from TiKV, we cannot distinguish that the key does not exist or the value of the key is empty. It is not a problem for TiDB but it is abnormal for a general KV database.

After the proto change, TiKV will set `not_found` to true if the key really doesn't exist. The returned `value` is not changed, so there is no compatibility issue when using an old TiDB with a new TiKV.

Additionally, I propose that [client-go](https://github.com/tikv/client-go) should change the interfaces to support empty value. After that, if TiDB will use client-go in the future, it should pay attention to treating an empty value as non-existence (in case TiDB wrote empty values before).

The added field name is called `not_found` (with negative meaning) instead of something like `key_exists` to keep the default meaning unchanged. Therefore, it is not a big problem when a new TiKV client connects to an old TiKV.

PTAL @disksing 

Signed-off-by: Yilin Chen <sticnarf@gmail.com>


proto/kvrpcpb.proto
```diff
@@ -125,6 +126,7 @@ message KvPair {
     KeyError error = 1;
     bytes key = 2;
     bytes value = 3;
+    bool not_found = 4;
```
disksing: Seems not useful.

sticnarf: I though BatchGet still returned non-existent keys. They are actually filtered. I'll remove it.

sticnarf: PTAL again. Thanks.

PR #440, opened 2019-08-13
tikvrpc: support batch split

disksing:


proto/kvrpcpb.proto
```diff
@@ -487,12 +487,14 @@ message MvccGetByStartTsResponse {
 message SplitRegionRequest {
     Context context = 1;
     bytes split_key = 2;
```
overvenus: Maybe we can change it `repeated` instead of adding a new filed.

disksing: em, seems not a good idea. If new tidb connects to an old tikv, tikv will silently split the region into 2 with no error.

disksing:
Ping @overvenus @zimulala

proto/kvrpcpb.proto
overvenus: Could you mark it deprecated?

disksing: good idea

disksing: done @overvenus

PR #443, opened 2019-08-15
kvrpcpb: add the large transaction proto

tiancaiamao:
Add a new `min_commit_ts` field in PrewriteRequest to support the large transaction proto,
and also add the LargeTxnHeartBeat proto to update `lock_ttl`.

@coocood


coocood:
We need an RPC method for heartbeat request.

And we can remove `Large` from the `LargeTxnHeartbeet`, just `TxnHeartBeat`.

proto/kvrpcpb.proto
```diff
@@ -176,12 +176,12 @@ message PrewriteRequest {
 }
 
 // Used to update the lock_ttl of a large transaction to prevent it from been killed.
-message LargeTxnHeartBeatRequest {
+message TxnHeartBeatRequest {
     bytes primary_lock = 1;
```
coocood: Need `Context context = 1;` like other requests.

proto/kvrpcpb.proto
```diff
@@ -176,12 +176,12 @@ message PrewriteRequest {
...
     uint64 advise_lock_ttl = 2;
 }
 
-message LargeTxnHeartBeatResponse {
+message TxnHeartBeatResponse {
     errorpb.Error region_error = 1;
```
coocood: Need `KeyError`

proto/kvrpcpb.proto
```diff
@@ -241,6 +255,31 @@ message BatchRollbackResponse {
...
+// Otherwise, returns the TTL information in addition to commit/rollback.
+// CheckTxnStatusRequest also pushes the minCommitTS forword for large transaction.
+//
+// In the future, TiDB may not store TTL in the secondary lock, and it calls this API to
+// get the TTL from the primary key.
+//
+message CheckTxnStatusRequest {
+    Context context = 1;
+    bytes primary_key = 2;
+    uint64 start_version = 3;
```
youjiali1995: Is it`primary_key`'s `start_version`? Does it need a `read_version` to update lock's `min_commit_ts`?

proto/kvrpcpb.proto
youjiali1995: Does it need a `lock_ts`?

proto/kvrpcpb.proto
coocood: No, it's current transaction's start_version.

youjiali1995: How can we get commit info if we don't know the txn's `start_ts`?

coocood: You are right, we need another ts.

tiancaiamao: Yes, we need a ts to push forward the minCommitTS, and I add a new field here:
https://github.com/pingcap/kvproto/pull/443/files#diff-93fb3a82615ef50b8b76520198e3d36fR270

This may be the (startTS + 1) of the checker, or the cached ts in TiDB, or the real ts from PD.

proto/kvrpcpb.proto
```diff
@@ -241,6 +258,32 @@ message BatchRollbackResponse {
     KeyError error = 2;
 }
 
+// CheckTxnStatusRequest checks the status of a transaction.
+// If the TTL of the transaction is exhausted, abort that transaction.
+// Otherwise, returns the TTL information in addition to commit/rollback.
+// CheckTxnStatusRequest also pushes the minCommitTS forword for large transaction.
```
MyonKeminta:
```suggestion
// CheckTxnStatusRequest also pushes the minCommitTS forward for large transaction.
```

proto/kvrpcpb.proto
```diff
@@ -241,6 +258,32 @@ message BatchRollbackResponse {
...
+//
+// In the future, TiDB may not store TTL in the secondary lock, and it calls this API to
+// get the TTL from the primary key.
```
MyonKeminta: May this behavior reduces the performance, when we can be sure that the lock is not expired? I think maybe it's better to keep a TTL in the secondary keys, and if secondary's TTL is expired, check the primary key's TTL instead.

proto/kvrpcpb.proto
```diff
@@ -170,6 +170,23 @@ message PrewriteRequest {
     uint64 txn_size = 8;
     // Use for pessimistic transaction, used to check if a conflict lock is already committed.
     uint64 for_update_ts = 9;
+    // If min_commit_ts > 0, this is large transaction proto, the final commit_ts
+    // would be infered from min_commit_ts.
+    uint64 min_commit_ts = 10;
+}
+
+// Used to update the lock_ttl of a large transaction to prevent it from been killed.
+message TxnHeartBeatRequest {
```
MyonKeminta: Could you please move these two new messages after `PrewriteResponse` ? I think it's better not to put these new messages between a Request and a Response

MyonKeminta:
Could you generate the Rust code, so we can start working on the TiKV side?

proto/kvrpcpb.proto
tiancaiamao: Keep the TTL on the secondary keys is also acceptable.
This change is not the final version.

MyonKeminta:
Could you please merge master? TiKV's master branch requires some newer commits in kvproto's master branch now.

coocood:
LGTM

PR #445, opened 2019-08-20
Make the package name and file name for deadlock.pb match (#442)

nrc:
Port of https://github.com/pingcap/kvproto/pull/442 to release-3.0 branch


BusyJay:
I think you need to regenerate rust code.

nrc:
> I think you need to regenerate rust code.

Added, thanks.

PR #450, opened 2019-08-28
backup: initial support backup

overvenus:
This PR adds basic proto to support backup. Restore is not included and will be proposed separately.


proto/backup.proto
```diff
@@ -0,0 +1,89 @@
...
+
+    bytes start_key = 3;
+    bytes end_key = 4;
+    uint64 start_version = 5;
+    uint64 end_version = 6;
+}
+
+message Schema {
+    bytes db = 1;
+    bytes table = 2;
```
zhangjinpeng1987: Is table contain how to create the table?

proto/backup.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message ClusterIDError {
```
zhangjinpeng1987: Should contain err message.

zhangjinpeng1987:
@BusyJay PTAL

proto/backup.proto
overvenus: Yes, it contains model.tableinfo. https://github.com/pingcap/parser/blob/8d44c03502f6b50e6111101c5504488d17f411b3/model/model.go#L194

proto/backup.proto
overvenus: The error message is included `Error`.

proto/backup.proto
```diff
@@ -0,0 +1,89 @@
...
+
+message BackupRequest {
+    uint64 cluster_id = 1;
+
+    bytes start_key = 2;
+    bytes end_key = 3;
+    uint64 start_version = 4;
+    uint64 end_version = 5;
+
+    string path = 6;
```
zhangjinpeng1987: Can this be remote url?

overvenus: Yes, this is a URL.

PR #451, opened 2019-08-29
import_kvpb: Strip the first 11 bytes from key in WriteEngine, specify it in OpenEngine

lance6716:
> This changes the OpenEngine request to include a key-prefix P, such that every write of (K, V) while the engine is open would become (P || K, V). Is this good enough for key-rewriting for BR restore?


lance6716:
PTAL @kennytm @BusyJay

kennytm:
PTAL @5kbpers. 

This changes the OpenEngine request to include a key-prefix *P*, such that every write of (*K*, *V*) while the engine is open would become (*P* || *K*, *V*). Is this good enough for key-rewriting for BR restore? 

(The strategy is the same as what I've described in the shared doc this morning; it is not the same as the original description in tikv/importer#5).

5kbpers:
Cool, LGTM

PR #453, opened 2019-09-06
Fix build

nrc:
The build broke because protobuf release 2.8.1 which Cargo picked for grpcio, however, this crate used a Git dep and `replace`d protobuf 2.8.0. By using exactly 2.8.0 here, we force Cargo to do the same for grpcio, then the replace takes effect.

PTAL @Hoverbear @hicqu

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ protobuf-codec = ["protobuf-build/grpcio-protobuf-codec", "raft-proto/protobuf-c
 prost-codec = ["prost", "prost-derive", "bytes", "lazy_static", "protobuf-build/grpcio-prost-codec", "raft-proto/prost-codec", "grpcio/prost-codec"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


nrc:
review ping @hicqu PTAL

Cargo.toml
```diff
@@ -15,7 +15,7 @@ protobuf-codec = ["protobuf-build/grpcio-protobuf-codec", "raft-proto/protobuf-c
 prost-codec = ["prost", "prost-derive", "bytes", "lazy_static", "protobuf-build/grpcio-prost-codec", "raft-proto/prost-codec", "grpcio/prost-codec"]
 
 [dependencies]
-protobuf = { git = "https://github.com/nrc/rust-protobuf", branch = "v2.8" }
+protobuf = "=2.8.0"
```
zhangjinpeng1987: duplicated `=` ?

nrc: No, the `=` tells Cargo to use exactly 2.8.0, and not to use 2.8.1, which doesn't match the replace

PR #458, opened 2019-09-12
tikvrpc: support batch split (#440)

disksing:


Cargo.toml
```diff
@@ -16,7 +16,7 @@ prost-codec = ["prost", "prost-derive", "bytes"]
 regenerate = []
 
 [dependencies]
-protobuf = { version = "2", optional = true }
+protobuf = { version = "=2.0.6", optional = true }
```
zimulala: Why change this version?

disksing: Without '=' it will use latest '2.x', which is '2.8.0' now and has compatibility problem.

PR #459, opened 2019-09-16
kvrpcpb: add a commit ts field to TxnHeartBeatResponse

tiancaiamao:
The [ttlManager goroutine](https://github.com/pingcap/tidb/pull/12177) may not know a txn is already committed/rollbacked and still sending the heartbeat request, so we need to add this commit ts field.

@lysu @coocood @MyonKeminta


MyonKeminta:
Couldn't you try to notify that goroutine within TiDB? Since `get_txn_commit_info` is a little heavy in TiKV

tiancaiamao:
> Since get_txn_commit_info is a little heavy in TiKV

But the code should almost never run to that function, so I guess the performance doesn't matter?

MyonKeminta:
When `txn_heart_beat` fails it returns a LockNotFound error. I think you can stop heartbeating when you receive this error.
If you really need the exact commit ts when heartbeating on a committed lock and that's impossible to implement within TiDB, this PR LGTM.

PR #455, opened 2019-09-09
Add current_ts field to Cleanup request

MyonKeminta:
This PR adds `current_ts` field to `CleanupRequest`. Inorder to optimize pessimistic transaction, we are going to implement TxnHeartBeat functionality on release-3.0 branch, and allow cleanup request check lock's ttl. It needs adding `current_ts` field to Cleanup request to tell TiKV which time to use to check the TTL. And inorder to allow gracefully rolling update from release-3.0 to newer version, this field is better also add to master branch. Now this PR adds the field to master branch, and will be cherry-picked to release-3.0 branch.l

PR #460, opened 2019-09-17
3.0: Add txn_heart_beat API and add current_ts to cleanup request

MyonKeminta:
This is used to allow refreshing TTL for pessimistic  transactions.

Comes from #443 and #455 .


MyonKeminta:
@tiancaiamao @coocood @youjiali1995 PTAL

coocood:
LGTM

PR #463, opened 2019-09-30
import_sstpb: added Download request

kennytm:
Added a "Download" gRPC request to copy an SST file from an external storage. It also includes option to perform prefix-rewrite and speed limiting.


kennytm:
PTAL @5kbpers @overvenus

proto/import_sstpb.proto
```diff
@@ -97,3 +101,28 @@ message CompactRequest {
...
+    //
+    //  new_key = new_key_prefix + old_key[old_key_prefix_length..]
+    uint32 old_key_prefix_length = 10;
+    bytes new_key_prefix = 11;
+
+    // The download speed limit (bytes/second). Set to 0 for unlimited speed.
+    uint64 speed_limit = 12;
+}
+
+message DownloadResponse {
```
overvenus: How does client know whether the RPC is succeed?

kennytm: If the RPC failed the gRPC response should be an error.

```go
	Download(ctx context.Context, in *DownloadRequest, opts ...grpc.CallOption) (*DownloadResponse, error)
	//                                                                        this would be non-nil ^~~~~
```

overvenus: I am not sure if it is enough to express errors like EpochNotMatch/RegionNotFound...

kennytm: @overvenus This request only copies a file from external_storage into local storage (plus key rewrite), and won't interact with PD. So `EpochNotMatch` and `RegionNotFound` should never happen.

overvenus: LGTM

PR #465, opened 2019-10-10
import_sstpb: abstract rewrite rule into its own message

kennytm:
This message type will be needed for other requests.

Also fixed some comments.


kennytm:
PTAL @5kbpers @overvenus 

To rewrite the table ID you replace the 10 byte prefix.

To rewrite the TS you replace the 8 byte suffix (encoded as `BITNOT(ts)` in big endian AFAIK).

kennytm:
PTAL again @5kbpers @overvenus

PR #466, opened 2019-10-10
Cherry-pick #462 and #464 into ov/engine 

birdstorm:
Cherry-pick critical PRs to support separated engine address for learner.


birdstorm:
@disksing can I merge this PR even though the CI reports `error: trait objects without an explicit `dyn` are deprecated`? 
I think it is due to CI using the latest rust version.

PR #467, opened 2019-10-15
improve backup checksum feature

NingLin-P:
Signed-off-by: linning <linningde25@gmail.com>
improve backup checksum feature


proto/backup.proto
```diff
@@ -39,17 +39,25 @@ message BackupMeta {
 
 message File {
     string name = 1;
-    uint32 crc32 = 2;
+    uint64 content_checksum = 2;
```
overvenus: Could you give it a self-described name?

3pointer: shoud we change it to `string`, if we use hash function like [sha-1](https://docs.rs/rust-crypto/0.2.36/crypto/sha1/index.html)

NingLin-P: like the used algorithm name?

5kbpers: prefer to use `bytes`

overvenus: Or maybe `bytes`?

NingLin-P: OK, I will use `bytes`

PR #469, opened 2019-10-17
import_sstpb: return the downloaded SST range and whether it is empty

kennytm:


kennytm:
PTAL @5kbpers @overvenus

PR #472, opened 2019-10-18
add LockType field into proto LockInfo

cfzjywxk:
add `LockType` in `LockInfo`  struct

default `0` means common lock,   '1' means pessimistic lock


coocood:
We can use `Op` as the type  https://github.com/pingcap/kvproto/blob/master/proto/kvrpcpb.proto#L137

cfzjywxk:
> We can use `Op` as the type https://github.com/pingcap/kvproto/blob/master/proto/kvrpcpb.proto#L137

@coocood 
for common lock, the op value is 2 not 0, do we need to make the `lockType` default value 0 ?

coocood:
We can just check if the Op is pessimistic lock

PR #473, opened 2019-10-18
Add new key error "commit_ts_expired"

MyonKeminta:
Large transactions will record the valid min_commit_ts on their primary key, and it can never be committed with a commit_ts less than the min_commit_ts. In this case we want to use min_commit_ts as the actual commit ts, but however we can't because it will cause key collision in write CF.

In our early implementation, commit should return an error when the commit_ts is less than min_commit_ts. This PR adds a new Error type to represent it.


MyonKeminta:
It's weird that so many go files are changed.....

sticnarf:
I have a question... If `CommitTsExpired` error is only used in `CommitResponse`, why not just add it to the `CommitResponse` instead of adding it to `KeyError`?

MyonKeminta:
@sticnarf It might not be a good idea. During a rolling update, when TiKV is upgraded but TiDB is not, a commit may fail and return the error in the new field of `CommitResponse`, but TiDB doesn't know the new field and think that it was committed successfully and continue committing secondaries.

MyonKeminta:
@sticnarf @coocood @tiancaiamao PTAL

tiancaiamao:
LGTM

PR #482, opened 2019-10-29
Add end_point_request_max_handle_duration in context

liufuyang:
To support push down TiDB's max execution time to TiKV's,
so that we can have a per-request max execution time support.

First step of https://github.com/tikv/tikv/issues/5712


proto/kvrpcpb.proto
```diff
@@ -84,6 +84,7 @@ message Context {
     bool scan_detail = 11; // true means return scan cf's detail
     bool replica_read = 12;
     repeated uint64 resolved_locks = 13;
+    uint64 end_point_request_max_handle_duration = 14;
```
breeswish: prefer to simply call it "max_handle_duration_ms" or something like that

lonng: @breeswish The coprocessor execution time contains `wait_time` and `handle_time`, I think it's better to `max_execution_duration_ms` is more appropriate.

liufuyang: Should I change it? `handle_time` is a bool though. 🤔

breeswish: @lonng Nice catch. `max_execution_duration_ms` +1

@liufuyang Actually the field `handle_time` means whether or not the handle time should be recorded.

PR #483, opened 2019-10-30
Allow CheckTxnStatusRequest specify rollback or not if txn not exist

MyonKeminta:
This PR adds new KeyError type `TxnNotFound` and added a new field `bool rollback_if_not_exist` to `CheckTxnStatusRequest`. When `rollback_if_not_exist` is false and `check_txn_status` finds no lock and  commit/rollback record, it will not write the rollback record and will return the TxnNotFound error.


tiancaiamao:
LGTM

PR #484, opened 2019-11-01
PessimisticLock: signed wait timeout

youjiali1995:
Signed-off-by: youjiali1995 <zlwgx1023@gmail.com>


coocood:
@sticnarf PTAL

coocood:
@MyonKeminta PTAL

coocood:
> Maybe better to comment for what negative value means?

+1
And comment for 0 value means default.

youjiali1995:
@coocood @sticnarf PTAL again. Thanks!

PR #489, opened 2019-11-12
add applied index in coprocessor request

hicqu:

So that TiKV can get snapshots on followers after it applys to `applied_index`, without any network commucation to regions' leader.


hicqu:
@Connor1996 description is updated. PTAL.

proto/coprocessor.proto
```diff
@@ -25,6 +25,8 @@ message Request {
     int64 tp = 2;
     bytes data = 3;
     repeated KeyRange ranges = 4;
+
+    uint64 applied_index = 5;
```
Connor1996: better to add a comment here

PR #491, opened 2019-11-13
kvrpcpb: record rollback reason in the CheckTxnStatus response

tiancaiamao:
When the CheckTxnStatus returns rollbacked, provide the rollback reason.

We need this information to debug https://github.com/pingcap/tidb/issues/13397

@coocood @youjiali1995  @MyonKeminta


proto/kvrpcpb.proto
```diff
@@ -293,6 +293,12 @@ message CheckTxnStatusRequest {
     bool rollback_if_not_exist = 6;
 }
 
+enum RollbackReason {
+    NotByMe = 0;
```
MyonKeminta: How about `AlreadyRolledBack`, since it also may because duplicated requests

tiancaiamao: I think it's fine because "Me" can also refer to "this request".

B.T.W, RolledBack or Rollbacked, how to spell

coocood: `NotByMe` does not look like a reason.

tiancaiamao: Or we use 'Default' to mean no information?
When the txn status is not rollbacked, we do not need a reason, but we still have to provide a value. @coocood

coocood: Or `NoReason` as default

coocood:
LGTM

proto/kvrpcpb.proto
tiancaiamao: Done

PR #493, opened 2019-11-15
move applied_index to kvrpcpb.Context

hicqu:


hicqu:
PTAL @zhangjinpeng1987  @5kbpers thanks!

proto/coprocessor.proto
```diff
@@ -25,10 +25,6 @@ message Request {
     int64 tp = 2;
     bytes data = 3;
     repeated KeyRange ranges = 4;
-
-    // After a region applys to `applied_index`, we can get a
-    // snapshot for the region even if the peer is follower.
-    uint64 applied_index = 5;
```
breeswish: do we need to reserve 5?

hicqu: It's new  added a few days before today. So I guess not.

PR #494, opened 2019-11-15
Add fields for Coprocessor cache

breeswish:
Rebase to master for https://github.com/pingcap/kvproto/pull/365


sticnarf:
I worry it's not a good idea to use raft log index as the cache key. It can be interfered by other non-write raft commands like read index.

Could we just remove cache key and use start_ts to decide whether cache is hit?

hicqu:
@sticnarf `read index` won't increase `apply index`. Only applying raft entries or snapshots can increase it.

sticnarf:
> @sticnarf `read index` won't increase `apply index`. Only applying raft entries or snapshots can increase it.

Got it. Thanks.

PR #495, opened 2019-11-18
proto/kvrpcpb: change the rollback_reason field  of CheckTxnStatusResponse to action

tiancaiamao:
For compatibility reasons, we need to record the `MinCommitTSPushed` information in the `CheckTxnStatus` response.

This commit reuse the previous `rollback_reason` field, since it's only in the master branch, this direct field change may no impose any compatibility issues.

@coocood @youjiali1995 @lysu


proto/kvrpcpb.proto
```diff
@@ -297,10 +297,11 @@ message CheckTxnStatusRequest {
     bool rollback_if_not_exist = 6;
 }
 
-enum RollbackReason {
-    NoReason = 0;
-    TTLExpire = 1;
-    LockNotExist = 2;
+enum Action {
+    Rollbacked = 0;
```
coocood: We already know if it is rollbacked by lock_ttl and commit_version.
The default value should be NoAction.

coocood:
LGTM

PR #497, opened 2019-11-20
support follower read with applied index

hicqu:


hicqu:
The RFC is merged. PTAL @BusyJay @5kbpers thanks!

PR #498, opened 2019-11-20
config: initial config protocol

overvenus:
This PR adds config protocol.


proto/configpb.proto
```diff
@@ -0,0 +1,97 @@
...
+    uint64 cluster_id = 1;
+    Error error = 2;
+    Version version = 3;
+    string config = 4;
+}
+
+message GetRequest {
+    uint64 cluster_id = 1;
+    Version version = 2;
+    bytes component_id = 4;
```
rleungx: should be `3`?

proto/configpb.proto
```diff
@@ -0,0 +1,97 @@
...
+option (gogoproto.unmarshaler_all) = true;
+option (rustproto.lite_runtime_all) = true;
+
+service Config {
+    rpc Create(CreateRequest) returns (CreateResponse) {}
+    rpc Get(GetRequest) returns (GetResponse) {}
+    rpc Update(UpdateRequest) returns (UpdateResponse) {}
+}
+
+message Error {
```
lonng: It's better to change the `Error` to `Status` like the `RocksDB` status.

proto/configpb.proto
overvenus: Tag 3 has been assigned to a new field component.

proto/configpb.proto
```diff
@@ -0,0 +1,98 @@
...
+    rpc Create(CreateRequest) returns (CreateResponse) {}
+    rpc Get(GetRequest) returns (GetResponse) {}
+    rpc Update(UpdateRequest) returns (UpdateResponse) {}
+}
+
+message Status {
+    enum StatusCode {
+        UNKNOWN = 0;
+        FAILED = 1;
+        STALE_VERSION = 3;
```
rleungx: should be `2`?

proto/configpb.proto
```diff
@@ -0,0 +1,99 @@
...
+    Version version = 2;
+    Component component = 3;
+    bytes component_id = 4;
+}
+
+message GetResponse {
+    uint64 cluster_id = 1;
+    Status status = 2;
+    Version version = 3;
+    string config = 4;
```
NingLin-P: Should we explicitly set  `config` in here and `CreateResponse` to `optional`?

overvenus: The configpb.proto uses proto3 syntax which all fields are `optional` by default and the `required` has been removed.

proto/configpb.proto
lonng: Should we add variant `NOT_CHANGE` which represents the configuration do not change?

proto/configpb.proto
```diff
@@ -0,0 +1,99 @@
...
+        STALE_VERSION = 3;
+    }
+    StatusCode code = 1;
+    string message = 2;
+    Version version = 3;
+}
+
+message Version {
+    uint64 local = 1;
+    uint64 global = 2;
```
disksing: Need some comments? Hard to know what `local` and `global` mean.

proto/configpb.proto
```diff
@@ -0,0 +1,99 @@
...
+}
+
+enum Component {
```
disksing: Maybe better to use string instead of enum.

proto/configpb.proto
```diff
@@ -0,0 +1,99 @@
...
+    UNKNOWN = 0;
+    PD = 1;
+    TiKV = 2;
+    TiDB = 3;
+}
+
+message Local {
+    bytes component_id = 1;
```
disksing: Seems strings is better.

overvenus: I am curious why it is better.

proto/configpb.proto
overvenus: I am curious why it is better.

proto/configpb.proto
disksing: I guess the component identifier should be some readable text?

proto/configpb.proto
disksing: Because PD/TiKV/TIDB (specific service names) introduce unnecessary conceptual dependencies.

proto/configpb.proto
```diff
@@ -60,8 +53,8 @@ message ConfigEntry {
 message CreateRequest {
     uint64 cluster_id = 1;
     Version version = 2;
-    bytes component_id = 3;
-    Component component = 4;
+    string component = 3;
```
lonng: I think we could keep in consistent with TiDB system table TIDB_CLUTER_INFO, use TYPE and ADDRESS to represent a cluster component node, which maybe make the user have a more understanding.

overvenus: It's not only for tidb but also for other services, I prefer to keep it generic.

proto/configpb.proto
```diff
@@ -0,0 +1,117 @@
...
+    rpc Get(GetRequest) returns (GetResponse) {}
+    rpc Update(UpdateRequest) returns (UpdateResponse) {}
+    rpc Delete(DeleteRequest) returns (DeleteResponse) {}
+}
+
+message Status {
+    enum StatusCode {
+        UNKNOWN = 0;
+        OK = 1;
+        STALE_VERSION = 2;
```
NingLin-P: We should change `STALE_VERSION ` to `WRONG_VERSION`, so we can handle both version too high or too low.
BTW should we also add a `version` field on DeleteResponse?

proto/configpb.proto
```diff
@@ -0,0 +1,120 @@
...
+
+service Config {
+    rpc Create(CreateRequest) returns (CreateResponse) {}
+    rpc Get(GetRequest) returns (GetResponse) {}
+    rpc Update(UpdateRequest) returns (UpdateResponse) {}
+    rpc Delete(DeleteRequest) returns (DeleteResponse) {}
+}
+
+message Status {
+    enum StatusCode {
```
NingLin-P: Using the name `StatusCode` will generate `StatusStatusCode` in rust, maybe `Code` is ok.

PR #499, opened 2019-11-20
add ALL to the enum LogLevel

lonng:
Signed-off-by: Lonng <heng@lonng.org>

Add a variant `ALL` which represents all levels to the enum `LogLevel`;


proto/diagnosticspb.proto
```diff
@@ -20,12 +20,13 @@ service Diagnostics {
 }
 
 enum LogLevel {
-	Debug = 0;
-	Info = 1;
-	Warn = 2;
-	Trace = 3;
-	Critical = 4;
-	Error = 5;
+	Unspecified = 0;
```
overvenus: Usually, proto3 names it `UNKNOWN `.

https://developers.google.com/protocol-buffers/docs/proto3#enum

lonng: @overvenus The variant represents all log level instead of some special level. So I think `Unspecified` is more appropriate.

PR #501, opened 2019-11-21
Add Green GC proto

MyonKeminta:
This PR adds proto for Green GC.


proto/kvrpcpb.proto
```diff
@@ -593,6 +593,45 @@ message UnsafeDestroyRangeResponse {
...
+    Context context = 1;
+    uint64 max_ts = 2;
+}
+
+message RemoveLockObserverResponse {
+    string error = 1;
+}
+
+message PhysicalScanLockRequest {
+    Context context = 1;
```
sticnarf: Do we really need it?

MyonKeminta: No. But since I'm afraid that in the future there might be more fields  in context that might be useful for these requests, I kept the context for all of them. I also kept context for `UnsafeDestroyRangeRequest`.

PR #503, opened 2019-11-29
Add Location and group id for Follower Replication

Fullstop000:
Part of https://github.com/pingcap/community/issues/86

Add a new type `Location` to represent the geoinformation of a store in `metapb.proto`
Add a new field `group_id` in `RaftMessage` to represent the Group of `from_peer` in Follower Replication 

PTAL @hicqu @BusyJay 


proto/metapb.proto
```diff
@@ -44,6 +51,8 @@ message Store {
     // Status address provides the HTTP service for external components
     string status_address = 7;
     string git_hash = 8;
+    // Indicates the location the store is at.
+    Location location = 9;
```
hicqu: I think there are 2 solution for this:
1 is configuring `group-id`s with stores, just like here. So leader can know the associated `group-id` for every store by asking PD.
2 is add `group-id` in `RaftMessage`, so leader can know followers' `group-id` after they send any messages to the leader.
PTAL @BusyJay

breeswish:
@nrc Rust build is broken, could you take a look?

nrc:
Whoops, let me send a quick PR...

nrc:
Build is fixed by https://github.com/pingcap/kvproto/pull/505

PR #507, opened 2019-12-04
backup: extend backup storage for S3

tennix:
Signed-off-by: tennix <ztennix@gmail.com>

Extend backup storage to use support extra storage backend to avoid encoding all parameters in the `path` field as URL format.

*Note:* I'll generate the code in another commit once the protobuf definition is approved to make the review easier.


proto/backup.proto
```diff
@@ -88,6 +88,28 @@ message BackupRequest {
...
     uint32 concurrency = 8;
+
+    oneof storage_backend {
+        Local local = 9;
+        S3 s3 = 10;
+    }
+}
+
+message Local {
+    string path = 1;
```
tennix: New client should use this `path` for local storage, and TiKV server will try to use `storage_backend` and fallback to the `path` field in `BackRequest`.

kennytm: Given [protobuf's backward-compatibility design](https://developers.google.com/protocol-buffers/docs/proto3#updating), you could turn that `oneof` into

```protobuf
oneof storage_backend {
    string local_path = 6;
    bool noop = 9;
    S3 s3 = 10;
}
```

and drop the `string path = 6` field.

This assumes we entirely abandon `path` and not use it to store something like `s3://$bucket/$prefix`.

tennix: Oh, I forget there is a `noop` storage backend for testing.
Shall we abandon the `path` entirely?

kennytm: I don't think there's any configuration which applies to Local but not other cloud storage (file permission can be easily `chmod`'ed afterwards) 🙃

proto/backup.proto
```diff
@@ -82,12 +82,30 @@ message BackupRequest {
...
     // The I/O rate limit for backup request.
     uint64 rate_limit = 7;
     // The concurrency for executing the backup request in every tikv node.
     uint32 concurrency = 8;
+
+    oneof storage_backend {
+        // The path where saves backup files.
+        string local_path = 6;
+	bool noop = 9;
+        S3 s3 = 10;
```
overvenus: Indentation

proto/backup.proto
```diff
@@ -82,12 +82,30 @@ message BackupRequest {
...
+
+    oneof storage_backend {
+        // The path where saves backup files.
+        string local_path = 6;
+        bool noop = 9;
+        S3 s3 = 10;
+    }
+}
+
+message S3 {
```
DanielZhangQD: No need for `provider`?

proto/backup.proto
```diff
@@ -82,12 +82,30 @@ message BackupRequest {
...
+    string endpoint = 1;
```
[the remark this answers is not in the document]
tennix: Yes, with this struct we don't have to encode the endpoint in the `path` as URL, so no need to escape `://`.

proto/backup.proto
tennix: Yes, we can treat non-AWS providers as custom providers, and enfore to provide endpoint.

DanielZhangQD: OK, by `enforce`, we can only make it clear in the document, right? We cannot enforce it in code without provider.

proto/backup.proto
```diff
@@ -82,12 +82,35 @@ message BackupRequest {
...
-    // The path where saves backup files.
-    string path = 6;
     // The I/O rate limit for backup request.
     uint64 rate_limit = 7;
     // The concurrency for executing the backup request in every tikv node.
     uint32 concurrency = 8;
+
+    oneof storage_backend {
+        // The path where saves backup files.
+        string local_path = 6;
```
overvenus: Could you add a message for `Local` and deprecate `string path = 6`?

proto/backup.proto
overvenus: @kennytm Since we add `Noop noop = 9`, I prefer deprecate `string path = 6`, and add a dedicate `message Local` too.

PR #510, opened 2019-12-10
backup: update sse to string

DanielZhangQD:
SSE configures the server-side encryption algorithm used when storing this object in S3 and the type should be string.

### DIFF
--- a/proto/backup.proto
+++ b/proto/backup.proto
@@ -118,7 +118,7 @@ message S3 {
     string prefix = 4;
     string storage_class = 5;
     // server side encryption
-    bool sse = 6;
+    string sse = 6;
     string acl = 7;
     string access_key = 8;
     string secret_access_key = 9;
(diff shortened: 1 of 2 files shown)


proto/backup.proto
```diff
@@ -118,7 +118,7 @@ message S3 {
     string prefix = 4;
     string storage_class = 5;
     // server side encryption
-    bool sse = 6;
+    string sse = 6;
```
kennytm: @overvenus does this `6` need to be changed to another tag?

overvenus: I think it's okay to keep `6` since the new `S3` is not used yet.  BTW, why does it need to be changed to `string`?

DanielZhangQD: It is a `string` instead of `bool`, please check detail in the [here](https://rclone.org/s3/#s3-server-side-encryption).

DanielZhangQD:
> LGTM. (Someone please send a PR to tikv master to update kvproto again after this is merged)

The kvproto for S3 is not updated in https://github.com/tikv/tikv/pull/6186, I think @yiwu-arbug should update the kvproto for S3.

kennytm:
Yes, tikv/tikv#6186 was merged before this PR, so a separated PR is need to update kvproto on TiKV master again. No need to cherry-pick to TiKV 3.1 though, I've included this in tikv/tikv#6207 already.

PR #513, opened 2019-12-10
Add is_raw_kv field to backup request and meta

MyonKeminta:
This PR adds is_raw_kv field to BackupRequest and BackupMeta. 
This is used to support raw kv mode of backup ( https://github.com/pingcap/br/issues/86 )


MyonKeminta:
@overvenus @3pointer PTAL

MyonKeminta:
@overvenus @3pointer PTAL again

PR #514, opened 2019-12-10
backup: add GCS proto

SunRunAway:
part of https://github.com/pingcap/br/issues/89

## GCS Support (1800 points)
Extend backup protobuf to support GCS


proto/backup.proto
```diff
@@ -124,6 +125,26 @@ message S3 {
     string secret_access_key = 9;
 }
 
+// GCS storage backend saves files into google cloud storage.
+message GCS {
+    string endpoint = 1;
```
DanielZhangQD: I think no endpoint for gcs.

proto/backup.proto
```diff
@@ -124,6 +125,26 @@ message S3 {
...
+    //	{
+    //		"entity": "project-owners-1052207218631",
+    //		"role": "OWNER",
+    //		"projectTeam": {
+    //			"projectNumber": "1052207218631",
+    //			"team": "owners"
+    //		}
+    //  }
+    // ]`
+    string acl_json = 5;
```
DanielZhangQD: This is object ACL, right?

proto/backup.proto
tennix: Do we need to specify the `endpoint` when using GCS? S3 storage backend needs this because there are S3 compatible storages.

proto/backup.proto
```diff
@@ -124,6 +125,26 @@ message S3 {
     string secret_access_key = 9;
 }
 
+// GCS storage backend saves files into google cloud storage.
+message GCS {
```
DanielZhangQD: Should include `projectID`, `location`?

proto/backup.proto
```diff
@@ -124,6 +125,26 @@ message S3 {
...
+    string endpoint = 1;
+    string bucket = 2;
+    string prefix = 3;
+    string storage_class = 4;
+    // acl json: like `[
```
tennix: I think this is a serialized json string. It doesn't have to be the format below. So I suggest modifying the comment. And also it's better to add a documentation link to these two fields since many people not knowing this.

proto/backup.proto
SunRunAway: See https://godoc.org/google.golang.org/api/option#WithEndpoint

proto/backup.proto
SunRunAway: Updated, with comments.

proto/backup.proto
SunRunAway: We do not have to. These are for creating bucket.

proto/backup.proto
SunRunAway: gcs client has this ability, see https://godoc.org/google.golang.org/api/option#WithEndpoint
Maybe useful when users need specify a IP to upload.

proto/backup.proto
SunRunAway: Updated.

PR #516, opened 2019-12-10
add new timestamp for restore rewrite rule

Connor1996:
Signed-off-by: Connor1996 <zbk602423539@gmail.com>


sticnarf:
The go generated file needs updating.

PR #517, opened 2019-12-10
backup: add force_path_style for s3

DanielZhangQD:
add force_path_style for s3


proto/backup.proto
```diff
@@ -122,6 +122,7 @@ message S3 {
     string acl = 7;
     string access_key = 8;
     string secret_access_key = 9;
+    string force_path_style = 10;
```
kennytm: This should either be a bool or enum

DanielZhangQD: done

PR #527, opened 2019-12-13
3.1: Add cf field to File (#526)

MyonKeminta:
Cherry-picks #526


MyonKeminta:
@3pointer @overvenus PTAl

PR #529, opened 2019-12-18
add pd endpoints in StoreHeartbeatResponse

hicqu:


hicqu:
It's redundant with dynamic configuration change. We can close it.

PR #530, opened 2019-12-23
add FollowerReadProxy

hicqu:
Add a new service `FollowerReadProxy`.

For follower read crossing multiple data centers, TiDBs can send requests to the service to get a timestamp and all committed indices for target regions.

Generally the service will run on TiKVs on the major data center.


BusyJay:
Why not just add a method to kv service?

hicqu:
@BusyJay not all TiKVs needs to register this service. By design, only TiKVs in the major data centers need this. And an another reason is we think seperate them can make code more clean.
Is there any advantage to merge them?

BusyJay:
Union them can lead to more unify setup and reduce boilerplate codes. I think it may not be easy to identify whether it's a major data center, or whether it will become a major data center in the future.

disksing:
Do we need to proceed with this pull request? cc @BusyJay @hicqu

disksing:
Closing for inactivity.

PR #531, opened 2019-12-25
add grpc gateway for update config

rleungx:
This PR uses the gRPC gateway for updating config.


proto/configpb.proto
```diff
@@ -12,7 +13,9 @@ option (rustproto.lite_runtime_all) = true;
 service Config {
     rpc Create(CreateRequest) returns (CreateResponse) {}
     rpc Get(GetRequest) returns (GetResponse) {}
-    rpc Update(UpdateRequest) returns (UpdateResponse) {}
+    rpc Update(UpdateRequest) returns (UpdateResponse) {
+        option (google.api.http) = { post: "/pd/api/v1/config" body: "*" };
```
overvenus:
```suggestion
        option (google.api.http) = { post: "/config" body: "*" };
```

I think just `/config` is enough, it's better to not leak PD and api version to protobuf definition

proto/configpb.proto
```diff
@@ -12,7 +13,9 @@ option (rustproto.lite_runtime_all) = true;
 service Config {
     rpc Create(CreateRequest) returns (CreateResponse) {}
     rpc Get(GetRequest) returns (GetResponse) {}
```
overvenus: Please add gateway for other methods too.

.circleci/config.yml
```diff
@@ -16,6 +16,9 @@ jobs:
             echo 'export PATH=$HOME/.local/bin:$PATH' >> $BASH_ENV
             curl -OL https://github.com/google/protobuf/releases/download/v$PROTOC_VERSION/protoc-$PROTOC_VERSION-linux-x86_64.zip
             unzip protoc-$PROTOC_VERSION-linux-x86_64.zip -d $HOME/.local
+      - run:
+          name: Install protoc-gen-grpc-gateway
+          command: go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
```
overvenus: Could you use a specific version?

proto/configpb.proto
rleungx: This PR is only for compatibility.

PR #535, opened 2020-01-02
coprocessor: add schema version for tiflash

lzmhhh123:
`schema_ver` is for TiFlash to check whether the schema is out of date.


proto/coprocessor.proto
```diff
@@ -31,6 +31,8 @@ message Request {
     // its last version matches this `cache_if_match_version`.
     bool is_cache_enabled = 5;
     uint64 cache_if_match_version = 6;
+    // schema_ver is for TiFlash to check whether schema is out of date.
```
zanmato1984: I suggest not to limit the usage of the schema version only for TiFlash. Maybe we can say it is for "any schema-ful storage to validate schema correctness if necessary".

PR #543, opened 2020-02-05
*: refine scripts

BusyJay:
- Use `go env GOPATH` to check environment variables
- Share common scripts
- Correct requirements in README


Makefile
```diff
@@ -8,17 +8,19 @@ all: go rust c++
 
 init:
 	mkdir -p $(CURDIR)/bin
-go: init
+check: init
+	$(CURDIR)/scripts/check.sh
+go: check
```
tiancaiamao: The chek script checks libprotoc=v3.8.x, but generating go code doesn't depend on libprotoc, is this dependency really reasonable?

tiancaiamao: I have to point out that the real tool version is controlled by the file `tools.json`.
It's maintained by the `retool` binary, which checks `tools.json` and downloads the correct go tools to generate the code. @BusyJay

BusyJay: I think go does depend on protoc:

https://github.com/pingcap/kvproto/blob/892c45fe83288c8dd62b8139e700014254f79a7a/generate_go.sh#L72

tiancaiamao:
LGTM

PR #544, opened 2020-02-07
backup: add backup ddl to meta

5kbpers:

Support to backup DDLs to BackupMeta.


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/pingcap/kvproto
 
 require (
-	github.com/gogo/protobuf v1.3.1
+	github.com/gogo/protobuf v0.0.0-20180717141946-636bf0302bc9
```
kennytm: Why?

MyonKeminta:
Is it necessary to update protobuf version?

5kbpers:
> Is it necessary to update protobuf version?

I use a different version of protoc and that updates it. Did not notice pushed the `go.mod`. 😓

MyonKeminta:
Emmm.... I think if the protoc version is correct, there should not be so much changes in go files...

proto/backup.proto
```diff
@@ -40,6 +40,9 @@ message BackupMeta {
     // backup data's range is represented by raw_ranges.
     bool is_raw_kv = 8;
     repeated RawRange raw_ranges = 9;
+
+    // In incremental backup, DDLs which are completed in (lastBackupTS, backupTS] will be stored here.
+    bytes ddls = 10;
```
overvenus: What's the actual format of these DDLs? And how about `repeated bytes ddls = 10`

5kbpers: it is an array of `model.Job` that unmarshalled in JSON format, I think `bytes` is good enough.

5kbpers:
PTAL

kennytm:
Please revert all changes to `*.pb.go` except `backup.pb.go`.

5kbpers:
> LGTM, but if there's a solution to pass the jobs structurally instead of JSON it would be even better.

What we need to store is `model.Job` so I think JSON is more convenient. 😅

PR #547, opened 2020-02-12
set last commit index

BusyJay:
Early apply can introduce mismatched commit index and apply index when not shutdown gracefully. This PR adds last commit index to recover from it.


proto/raft_serverpb.proto
```diff
@@ -72,6 +72,7 @@ message RaftLocalState {
 
 message RaftApplyState {
     uint64 applied_index = 1;
+    uint64 last_commit_index = 3;
```
Little-Wallace: Don't we need a default value for it?

PR #550, opened 2020-02-13
learner need load merge target

gengliqi:
Signed-off-by: Liqi Geng <gengliqiii@gmail.com>

A learner should load merge target if another overlapped region wants to apply snapshot or create a peer.
So add a StatusCmdType to handle this.

For https://github.com/tikv/tikv/pull/6598


proto/raft_serverpb.proto
```diff
@@ -22,6 +22,7 @@ message RaftMessage {
     bytes end_key = 8;
     // If it has value, to_peer should be removed if merge is never going to complete.
     metapb.Region merge_target = 9;
+    MergeMsg merge = 10;
```
BusyJay: How about using StatusCommand in RaftCmdRequest?

proto/raft_cmdpb.proto
```diff
@@ -258,6 +258,8 @@ enum StatusCmdType {
     InvalidStatus = 0;
     RegionLeader = 1;
     RegionDetail = 2;
+    RegionMergeTarget = 3;
+    RegionWakeUp = 4;
```
Connor1996: what's used for

gengliqi: RegionMergeTarget used for learner peer load merge target.
RegionWakeUp used for waking up the hibernate region.

BusyJay: How about just keeping `RegionWakeUp`?

PR #553, opened 2020-02-14
Change physical scan lock API to non-streaming

MyonKeminta:
This PR means to solve the issue that the streaming API costs TiDB too much memory.
Since Green GC is not enabled in any released versions, so I think it's ok to change it directly.


sticnarf:
Should the response have a field indicating if there are more locks?

MyonKeminta:
I think it can be inferred from whether the number returned locks is equals to `limit`, like how the old ScanLock API does.

MyonKeminta:
I think it's better to merge this after a brief test

PR #554, opened 2020-02-14
learner need load merge target on release-3.1

gengliqi:
Signed-off-by: Liqi Geng <gengliqiii@gmail.com>

A learner should load merge target if another overlapped region wants to apply snapshot or create a peer.
So add a StatusCmdType to handle this.


5kbpers:
Please run `make rust` to generate the code of rust.

PR #565, opened 2020-02-27
add op_check_not_exists

lysu:
just like `op_insert` is "insert should not exists", we need another "delete should not exists".

https://github.com/pingcap/tidb/pull/14968

---
This change is [Reviewable](https://reviewable.io/reviews/pingcap/kvproto/565)


proto/kvrpcpb.proto
```diff
@@ -163,6 +163,7 @@ enum Op {
     // insert operation has a constraint that key should not exist before.
     Insert = 4;
     PessimisticLock = 5;
+    CheckExists = 6;
```
coocood: Have you considered `CheckNotExists`?

lysu: yes, it should be CheckNotExists, my fault- -

coocood:
LGTM

lysu:
hi @coocood, I found a but last night, we also need write key, nil when key not exists.

to avoid txnNotFoundError when `begin; insert record1;delete record1; commit`(we need prewrite something to help commit can works, maybe better way is check prewrite type and avoid commit phase, but that's more complex),

lysu:
@jackysp  PTAL if free, kvproto need two approve, then we can continue work on tidb/tikv pr

lysu:
now tidb can ensure never pass checkNotExists key into commit phase

coocood:
LGTM

jackysp:
@coocood please approve the pr.

proto/kvrpcpb.proto
```diff
@@ -163,6 +163,7 @@ enum Op {
     // insert operation has a constraint that key should not exist before.
     Insert = 4;
     PessimisticLock = 5;
+    CheckNotExists = 6;
```
youjiali1995: Why use the third person singular?😂

PR #566, opened 2020-03-04
encryption: add encryption protobuf messages 

overvenus:
Add encryption protobuf messages.

Note: this PR also replaces nrc/rusr-protobuf with tikv/rust-protobuf as the former does not support map fields currently https://github.com/nrc/rust-protobuf/pull/1


overvenus:
Seems protobuf-build does not support prost. Would you mind take a look? @nrc

nrc:
> Seems protobuf-build does not support prost. Would you mind take a look? @nrc

I'm on vacation this week, but I'll take a look on Monday. (I have merged nrc/rust-protobuf#1).

overvenus:
Have a nice vacation!

Map is supported by https://github.com/tikv/protobuf-build/pull/35

overvenus:
> Note: this PR also replaces nrc/rusr-protobuf with tikv/rust-protobuf as the former does not support map fields currently nrc/rust-protobuf#1

> (I have merged nrc/rust-protobuf#1).

FYI, can not switch to nrc/rust-protobuf as the package name is changed to nrc-protobuf, and can not replace protobuf any more.

PR #572, opened 2020-03-12
add target file for SearchLogRequest

Fullstop000:
Add target file option for `SearchLogRequest` since we've separated normal logs and slow logs
Part of https://github.com/tikv/tikv/issues/7069


Fullstop000:
@wish PTAL

PR #574, opened 2020-03-15
Remove 'All' type in SearchLogRequest

Fullstop000:


Fullstop000:
@lonng @breeswish PTAL

PR #575, opened 2020-03-16
import_sstpb, backuppb: add download error and tiflash replica count

5kbpers:

1. Add `error` field for distinguishing the error of the request with the error of gRPC.
2. Add tiflash replica count to backupmeta for recover the replica of tiflash


proto/import_sstpb.proto
```diff
@@ -143,6 +143,12 @@ message DownloadRequest {
     backup.StorageBackend storage_backend = 14;
 }
 
+// For now it is just used for distinguishing the error of the request with the error 
+// of gRPC, add more concrete types if it is necessary later.
+message Error {
+    string message = 1;
```
overvenus: Could you include errorpb.Error here?

5kbpers: I think it is not necessary, currently it is just used to return the error of `DownloadSST`.

PR #579, opened 2020-03-17
config: introduce a new method `GetAll`

qw4990:
Introduce a new method `GetAll` to let `TiDB` get all configs from `PD` and show them to the user.


pkg/cdcpb/cdcpb.pb.go
```diff
@@ -66,7 +66,7 @@ func (x Event_LogType) String() string {
 	return proto.EnumName(Event_LogType_name, int32(x))
 }
 func (Event_LogType) EnumDescriptor() ([]byte, []int) {
-	return fileDescriptor_cdcpb_12d02f2d7273b847, []int{2, 0}
+	return fileDescriptor_cdcpb_abb6ec925162558f, []int{2, 0}
```
rleungx: Why does this file change?

qw4990: It may be my environment's problem or someone changed the `cdcpb.proto` but not generated the `cdcpb.pb.go` correctly.
On the latest master branch, after running `make go`, these files also changed.
Can you reproduce this problem on the master branch in your computer? @rleungx

PR #581, opened 2020-03-20
Add end_key_exclude field to SstMeta and distinguish restoring of RawKv

MyonKeminta:
See https://github.com/tikv/tikv/issues/7163


MyonKeminta:
@kennytm  It's generated. Should I remove those changes?

kennytm:
Yeah please revert it. gogoprotoc creates a random name every time it is called, and thus all `*.pb.go` will be changed even if the `*.proto` is unchanged.

PR #584, opened 2020-03-23
pdpb: add replicate_mode

disksing:

See PD issue https://github.com/pingcap/pd/issues/2272 for detail.


proto/replicate_mode.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+package replicate_mode;
+
+// The replicate status sync from PD to TiKV.
+message ReplicateStatus {
+    enum Mode {
+        MAJORITY = 0;  // the standard mode. Replicate logs to majority peer.
+        MS_AUTOSYNC = 1; // master-slave mode. Replicate logs among 2 DCs.
```
BusyJay: How about "PS_AUTO"? "PS" stands for primary and secondary. Or "Active-Active".

proto/replicate_mode.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// The status of ms-autosync mode.
+message MSAutoSync {
+    enum State {
+        SYNC = 0; // raft logs need to sync between different DCs
+        ASYNC = 1; // raft logs need to sync to majority peers
+        SYNC_RECOVER = 2; // switching from ASYNC to SYNC mode
+    }
+    string label = 1; // the key of the label that used for distinguish different DC.
```
BusyJay: How about "label_key"?

proto/replicate_mode.proto
```diff
@@ -0,0 +1,35 @@
...
+    State state = 2;
+    uint64 recover_id = 3; // Unique ID for a recovery
+    int32 wait_sync_timeout_hint = 4; // Duration to wait before switching to SYNC by force
```
BusyJay: Should mention its unit.

proto/replicate_mode.proto
disksing: seems PS is harder for reader to guess its meaning. I'm thinking maybe just AUTOSYNC

proto/replicate_mode.proto
disksing: good idea

proto/replicate_mode.proto
disksing: good idea

proto/replicate_mode.proto
```diff
@@ -0,0 +1,35 @@
...
+
+// The replicate status sync from TiKV to PD.
+message RegionReplicateStatus {
+    enum State {
+        MAJORITY = 0; // Logs sync to majority peers
+        INTEGRITY_OVER_LABEL = 1; // Logs sync to different DCs
+    }
+    State state = 1;
+    uint64 recover_id = 2; // Unique ID for a recovery
+}
```
HunDunDM: Need new blank line？

disksing: good idea

disksing:
PTAL @BusyJay @nolouch @HunDunDM

proto/replicate_mode.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+package replicate_mode;
+
+// The replicate status sync from PD to TiKV.
+message ReplicateStatus {
```
BusyJay, line 5: Maybe we need to name it as `ReplicationStatus`?

disksing: I think it ok. You can send a PR.

PR #585, opened 2020-03-23
add trace spans to response

Renkai:
Add a field to `coprocessor.Response` which will be used in
https://github.com/tikv/tikv/pull/7049


disksing:
Hi @Renkai , do we want to continue this feature?

Renkai:
@disksing I think it's been too long and I almost lost all context, so it looks better to archive this pr.

PR #587, opened 2020-03-26
Add some documentation to the kv protos

nrc:
Moves around the messages in kvproto.md (second commit) and improves documentation (first and third commits). There are no material changes to any protos.

PTAL @overvenus @hicqu

### DIFF
--- a/proto/kvrpcpb.proto
+++ b/proto/kvrpcpb.proto
@@ -13,6 +13,9 @@ option (rustproto.lite_runtime_all) = true;
 
 option java_package = "org.tikv.kvproto";
 
+// This proto file defines requests, responses, and helper messages for KV and raw
+// APIs of TiKV (see tikvpb.proto).
+
 message LockInfo {
     bytes primary_lock = 1;
     uint64 lock_version = 2;
--- a/proto/kvrpcpb.proto
+++ b/proto/kvrpcpb.proto
@@ -18,6 +18,8 @@ option java_package = "org.tikv.kvproto";
 
 // Transactional commands.
 
+// A transactional get command. Lookup a value for `key` in the transaction with
+// starting timestamp = `version`.
 message GetRequest {
     Context context = 1;
     bytes key = 2;
@@ -25,17 +27,26 @@ message GetRequest {
 }
 
 message GetResponse {
+    // A region error indicates that the request was sent to the wrong TiKV node
+    // (or other, similar errors).
     errorpb.Error region_error = 1;
+    // A value could not be retrieved due to the state of the database for the requested key.
     KeyError error = 2;
+    // A successful result.
     bytes value = 3;
+    // True if the key does not exist in the database.
     bool not_found = 4;
 }
 
+// Scan fetches values for a range of keys; it is part of the transaction with
+// starting timestamp = `version`.
 message ScanRequest {
     Context context = 1;
     bytes start_key = 2;
+    // The maximum number of results to return.
     uint32 limit = 3;
     uint64 version = 4;
+    // Return only the keys found by scanning, not their values.
     bool key_only = 5;
     bool reverse = 6;
     // For compatibility, when scanning forward, the range to scan is [start_key, end_key), where start_key < end_key;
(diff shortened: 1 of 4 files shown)


proto/kvrpcpb.proto
```diff
@@ -401,6 +322,24 @@ message GCResponse {
     KeyError error = 2;
 }
 
+// Delete a range of data from TiKV. Similar to a scan, but deleting a range, rather than reading.
+message DeleteRangeRequest {
```
youjiali1995: @MyonKeminta PTAL. I remember this command is not recommended for use.

proto/kvrpcpb.proto
```diff
@@ -652,3 +530,212 @@ message ReadIndexResponse{
...
+    // snapshot for the region even if the peer is a follower.
+    uint64 applied_index = 15;
+}
+
+message LockInfo {
+    bytes primary_lock = 1;
+    uint64 lock_version = 2;
+    bytes key = 3;
+    uint64 lock_ttl = 4;
+    // How many keys this transaction involved.
```
youjiali1995:
```suggestion
    // How many keys this transaction involves in each region.
```

proto/kvrpcpb.proto
MyonKeminta: You are right. This is not recommended to use, which I think should be mentioned in the comment. By the way, it's not similar to scan, since it may invoke RocksDB's `delete_files_in_range` or `delete_range` to delete the data, and it will not consider constraints of transaction.

nrc:
Comments addressed. @youjiali1995 @MyonKeminta PTAL

nrc:
> LGTM. Have you run make to ensure no generated file is updated?

I had not and it seems the Go files changed. Good catch!

PTAL @youjiali1995 @MyonKeminta

proto/tikvpb.proto
```diff
@@ -46,30 +46,32 @@ service Tikv {
     rpc RawDeleteRange(kvrpcpb.RawDeleteRangeRequest) returns (kvrpcpb.RawDeleteRangeResponse) {}
     rpc RawBatchScan(kvrpcpb.RawBatchScanRequest) returns (kvrpcpb.RawBatchScanResponse) {}
 
-    // Store commands (to the whole tikv but not a certain region)
+    // Store commands (sent to a whole TiKV cluster, rather than a certain region).
```
MyonKeminta, line 49: I think these interfaces should be for a TiKV instance rather than the whole cluster.

proto/tikvpb.proto
```diff
@@ -46,30 +46,32 @@ service Tikv {
...
     rpc CoprocessorStream(coprocessor.Request)  returns (stream coprocessor.Response) {}
 
-    // Raft commands (tikv <-> tikv).
+    // Raft commands (sent between TiKV nodes).
     rpc Raft(stream raft_serverpb.RaftMessage) returns (raft_serverpb.Done) {}
     rpc BatchRaft(stream BatchRaftMessage) returns (raft_serverpb.Done) {}
     rpc Snapshot(stream raft_serverpb.SnapshotChunk) returns (raft_serverpb.Done) {}
 
-    // Region commands.
+    // Region commands (sent from PD to a TiKV node).
```
MyonKeminta, line 66: It seems not so. I'm not sure about `SplitRegion`, but `ReadIndex` is not from PD but from Raft followers.

proto/kvrpcpb.proto
```diff
@@ -646,6 +510,21 @@ message PhysicalScanLockResponse {
     repeated LockInfo locks = 2;
 }
 
+// Region commands (sent from PD to a TiKV node).
```
MyonKeminta, line 513: Here it should be changed too.

proto/tikvpb.proto
nrc: AIUI, each message is sent to one instance, but TiDB sends a message to every TiKV instance in the cluster

proto/tikvpb.proto
```diff
@@ -54,31 +54,34 @@ service Tikv {
...
-    // Raft commands (tikv <-> tikv).
+    // Raft commands (sent between TiKV nodes).
     rpc Raft(stream raft_serverpb.RaftMessage) returns (raft_serverpb.Done) {}
     rpc BatchRaft(stream BatchRaftMessage) returns (raft_serverpb.Done) {}
     rpc Snapshot(stream raft_serverpb.SnapshotChunk) returns (raft_serverpb.Done) {}
 
-    // Region commands.
+    // Sent from PD to a TiKV node.
     rpc SplitRegion (kvrpcpb.SplitRegionRequest) returns (kvrpcpb.SplitRegionResponse) {}
+    // Sent from TiFlash to a TiKV node.
```
youjiali1995, line 77: Not only TiFlash but also TiKV.

proto/kvrpcpb.proto
```diff
@@ -277,91 +147,110 @@ message PessimisticRollbackResponse {
     repeated KeyError errors = 2;
 }
 
-message CommitRequest {
-    reserved 5;
-    reserved "binlog";
+// Used to update the lock_ttl of a large transaction to prevent it from been killed.
```
MyonKeminta, line 150: It's also used in pessimistic transactions, not only large transactions.

proto/tikvpb.proto
```diff
@@ -54,31 +54,34 @@ service Tikv {
...
     rpc BatchCoprocessor(coprocessor.BatchRequest) returns (stream coprocessor.BatchResponse) {}
 
-    // Raft commands (tikv <-> tikv).
+    // Raft commands (sent between TiKV nodes).
     rpc Raft(stream raft_serverpb.RaftMessage) returns (raft_serverpb.Done) {}
     rpc BatchRaft(stream BatchRaftMessage) returns (raft_serverpb.Done) {}
     rpc Snapshot(stream raft_serverpb.SnapshotChunk) returns (raft_serverpb.Done) {}
 
-    // Region commands.
+    // Sent from PD to a TiKV node.
```
MyonKeminta, line 75: TiDB may also use this API. I suggest avoid mentioning what may invoke this API here.

nrc: I think it is useful to have the info, e.g., I did not know that TiDB could send SplitRegion :-) I've updated the comment

nrc:
updated, PTAL @MyonKeminta

PR #592, opened 2020-03-31
Support BR encryption

yiwu-arbug:
Add encryption related config to backup request and sst_import request.


proto/import_sstpb.proto
```diff
@@ -72,6 +73,9 @@ message SSTMeta {
     string cf_name = 5;
     uint64 region_id = 6;
     metapb.RegionEpoch region_epoch = 7;
+
+    // IV to be used if encryption is enabled.
+    bytes iv = 8;
```
yiwu-arbug, line 78: probably also need to add `sha256` field.

yiwu-arbug: no need to.

yiwu-arbug:
ready for review

proto/import_sstpb.proto
```diff
@@ -141,6 +145,8 @@ message DownloadRequest {
     RewriteRule rewrite_rule = 13 [(gogoproto.nullable) = false];
 
     backup.StorageBackend storage_backend = 14;
+
+    encryptionpb.EncryptionConfig encryption_config = 15;
```
kennytm, line 149: Is this needed for Upload and Ingest requests?

proto/backup.proto
```diff
@@ -43,6 +45,9 @@ message BackupMeta {
 
     // In incremental backup, DDLs which are completed in (lastBackupTS, backupTS] will be stored here.
     bytes ddls = 10;
+
+    // Encryption method and key.
+    encryptionpb.EncryptionConfig encryption = 11;
```
kennytm, line 50: what is the purpose of this config here?

if the backupmeta is in plaintext the key will be exposed. if the backupmeta is encrypted this field is redundant.

proto/import_sstpb.proto
yiwu-arbug: AFAIK BR only use the download API. Is it true?

proto/backup.proto
yiwu-arbug: backupmeta is encrypted using aes256-gcm (unconfigurable). data is encrypted using the config provided by user and stored here. I'm still debating with myself whether the two encryption should have different key - as they may be of different key length, and truncating a key doesn't sound right.

I figure I may need some document, as the implementation I got is not super straightforward. I'll send a RFC soon.

proto/import_sstpb.proto
kennytm: BR uses both Download and Ingest.

Upload is used by TiKV Importer at the moment.

yiwu-arbug: Just to confirm, `download` is for pulling files from storage,  `ingest` is to ask tikv to ingest the downloaded file, is that correct? In that case only `download` needs the config to decrypt the files. And this PR is for BR only.

kennytm: i see.

proto/backup.proto
kennytm: Given that we're going to use HKDF, please change this to 

```suggestion
     encryptionpb.EncryptionMethod encryption_method = 11; 
```

yiwu-arbug:
We will be using S3 server-side encryption to support BR encryption instead. Not moving forward for this one for now.

PR #596, opened 2020-04-07
Add PD API to support setting limit to GC safepoint

MyonKeminta:
Support getting and setting limit to GC safepoint, which makes it possible for other services to block GC.


proto/pdpb.proto
```diff
@@ -558,3 +562,25 @@ message GetOperatorResponse {
     OperatorStatus status = 4;
     bytes kind = 5;
 }
+
+message SetGCSafePointLimitRequest {
+    RequestHeader header = 1;
+    // UUID v4
+    bytes uuid = 2;
```
overvenus, line 569: I thought you want to change `id` (with the form of "servicename/uuid")?

proto/pdpb.proto
```diff
@@ -558,3 +562,25 @@ message GetOperatorResponse {
...
+}
+
+message GetGCSafePointLimitRequest {
+    RequestHeader header = 1;
+}
+
+message GetGCSafePointLimitResponse {
+    RequestHeader header = 1;
+    uint64 safe_point = 2;
+    uint64 min_safe_point_limit = 3;
```
overvenus, line 585: Why not extend the `GetGCSafePointResponse` instead?

nolouch:
PTAL @shafreeck

proto/pdpb.proto
MyonKeminta: Oh, sorry. I misunderstood the document here.

proto/pdpb.proto
```diff
@@ -558,3 +561,15 @@ message GetOperatorResponse {
     OperatorStatus status = 4;
     bytes kind = 5;
 }
+
+message SetGCSafePointLimitRequest {
+    RequestHeader header = 1;
+    // service_name/uuid_v4
+    bytes id = 2;
```
shafreeck, line 568: How about adding the `id` and `ttl` field in the `UpdateGCSafepoint` method, then it is not necessary to add a new method.

proto/pdpb.proto
```diff
@@ -498,6 +500,7 @@ message GetGCSafePointResponse {
     ResponseHeader header = 1;
 
     uint64 safe_point = 2;
+    uint64 min_safe_point_limit = 3;
```
shafreeck, line 503: I think the returned `safe_point` is always the minimum value, so `min_safe_point_limit ` is not necessary.

MyonKeminta: TiDB need to check how far it can advance the safepoint before it actually advance it, because it need to do `DeleteRange` step with the limited safepoint before sending its safepoint to PD.

MyonKeminta:
Use https://github.com/pingcap/kvproto/pull/603 instead. Closed.

PR #597, opened 2020-04-07
*: update name and script

BusyJay:
- Update go script to remove non-existing directory
- Update replication mode name to match conventions


BusyJay:
Comments are also adjusted so that they can appear in the generated sources.

PR #602, opened 2020-04-13
sst_importer: add write streaming request

3pointer:


proto/import_sstpb.proto
```diff
@@ -175,3 +178,29 @@ message SetDownloadSpeedLimitRequest {
 
 message SetDownloadSpeedLimitResponse {
 }
+
+message Mutation {
+    enum OP {
+        Put = 0;
+    }
+    OP op = 1;
```
kennytm, line 186: I don't think this is necessary for us.

```suggestion
```

proto/import_sstpb.proto
```diff
@@ -45,6 +45,9 @@ service ImportSST {
     // Download an SST file from an external storage, and performs key-rewrite
     // after downloading.
     rpc Download(DownloadRequest) returns (DownloadResponse) {}
+
+    // Open a write stream to generate sst file
```
kennytm, line 49:
```suggestion
    // Open a write stream to generate sst files
```

PR #604, opened 2020-04-17
add task_id to kv context

sticnarf:
`task_id` is added to let the client provide the information that multiple requests are of the same task. Currently it helps yatp to schedule tasks better. But it may be useful for other cases (maybe tracing) in the future.


youjiali1995:
LGTM. Should it run `make go` again?

BusyJay:
> Should it run make go again?

I did.

PR #607, opened 2020-04-22
Add sse_kms_key_id for S3 backup

yiwu-arbug:
Add `sse_kms_key_id` field to S3 config. This is to allow using an owned KMS key when having S3 server-side encryption.

Signed-off-by: Yi Wu <yiwu@pingcap.com>


yiwu-arbug:
> LGTM.
> 
> Should this be cherry-picked to 3.1?

no need right now.

yiwu-arbug:
But I do want to catch up with 4.0-rc.2 branch cut.

PR #622, opened 2020-05-21
Add proto for CDC to support large txn

MyonKeminta:
This PR adds proto for CDC to support push resolved_ts according to large transaction's min_commit_ts.


proto/cdcpb.proto
```diff
@@ -94,7 +118,17 @@ message ChangeDataRequest {
     bytes end_key = 6;
     // Used for CDC to identify events corresponding to different requests.
     uint64 request_id = 7;
+
     kvrpcpb.ExtraOp extra_op = 8;
+
+    oneof request {
+        // A normal request that trying to register change data feed on a region.
+        Register register = 9;
```
overvenus, line 126: Why do we need it?

MyonKeminta: I think there may be more different types of request so I use `oneof` field in the request. The `register` represents the original request that registers changefeed to a region.

MyonKeminta:
@overvenus PTAL again, thanks

PR #624, opened 2020-05-31
new proto files for minitrace

Renkai:
new protocol for https://github.com/tikv/tikv/pull/7781


proto/span.proto
```diff
@@ -0,0 +1,43 @@
...
+        Parent parent = 2;
+        Continue continue = 3;
+    }
+}
+
+
+message Span {
+    uint64 id = 1;
+    Link link = 2;
+    uint64 begin_cycles = 3;
```
breeswish, line 40: counted by cycles instead of ns?

proto/span.proto
```diff
@@ -0,0 +1,43 @@
...
+    HandleChecksum = 6;
+    HandleDag = 7;
+    HandleBatchDag = 8;
+    HandleAnalyze = 9;
+    HandleCached = 10;
+};
+
+message SpanSet {
+    uint64 start_time_ns = 1;
+    uint64 cycles_per_sec = 2;
```
breeswish, line 20: why not convert to ns?

proto/coprocessor.proto
```diff
@@ -6,6 +6,7 @@ import "kvrpcpb.proto";
 import "gogoproto/gogo.proto";
 import "rustproto.proto";
 import "metapb.proto";
+import "span.proto";
```
breeswish, line 9: The span might be shared with TiDB as well. Maybe tipb is a better place?

proto/coprocessor.proto
```diff
@@ -34,6 +35,7 @@ message Request {
     uint64 cache_if_match_version = 6;
     // Any schema-ful storage to validate schema correctness if necessary.
     int64 schema_ver = 8;
+    bool enable_trace = 9;
```
breeswish, line 38: `is_trace_enabled`

proto/span.proto
```diff
@@ -0,0 +1,43 @@
...
+    CoprRequest = 1;
+    Scheduled = 2;
+    Snapshot = 3;
+    HandleRequest = 4;
+    HandleUnaryFixture = 5;
+    HandleChecksum = 6;
+    HandleDag = 7;
+    HandleBatchDag = 8;
+    HandleAnalyze = 9;
+    HandleCached = 10;
```
breeswish, line 15: I think these event names needs further discussion. For now, I prefer to reserve these simpliest ones:

```
TiKVCoprHandleRequest
TiKVCoprScheduleTask
TiKVCoprGetSnapshot
TiKVCoprExecuteDAGRunner
TiKVCoprExecuteBatchDAGRunner
```

proto/span.proto
zhongzc: We leave the conversion work to the visualization part since heavy div jobs are needed.

proto/coprocessor.proto
BusyJay: I suggest to make minitrace a standalone project and keep the proto inside the project.

breeswish: The events are business dependent (as you can see from the enum variants) and might not be a good place to put it in minitrace.

BusyJay: You can define `Event event = 5` as `uint64 event = 5` and keep Event enum definition inside kvproto. kvproto is supposed to be used by TiKV, should not depend on tipb.

proto/span.proto
BusyJay: I think these events should be put inside tipb repos. There is no such `Dag` or `BatchDag` inside TiKV level. It's more tidb-related.

proto/coprocessor.proto
breeswish: Good idea!

proto/span.proto
breeswish: The time cost of an integer div each span might not be so notable and worth to have a try

zhongzc: The benchmark result leads me to make this decision. An integer div op cost more than 10ns.

PR #626, opened 2020-06-08
add sample_step option on ScanRequest.

coocood:
In order to support fast analyze on clustered index.


proto/kvrpcpb.proto
```diff
@@ -52,6 +52,8 @@ message ScanRequest {
     // For compatibility, when scanning forward, the range to scan is [start_key, end_key), where start_key < end_key;
     // and when scanning backward, it scans [end_key, start_key) in descending order, where end_key < start_key.
     bytes end_key = 7;
+    // If sample_step > 0, skips 'sample_step' number of keys after each returned key.
+    uint32 sample_step = 8;
```
MyonKeminta, line 56: I think semantically it's better to skip `sample_step - 1` keys and pick one keys.

PR #627, opened 2020-06-08
cdc: optionally output old value

5kbpers:

* Define `ExtraOp`
* Add `ExtraOp` option to `ChangeDataRequest`
* Add `old_value` to `Event`


proto/kvrpcpb.proto
```diff
@@ -858,3 +858,9 @@ message KeyRange {
     bytes start_key = 1;
     bytes end_key = 2;
 }
+
+enum ExtraRead {
+    Noop = 0;
+    Deleted = 1;
+    Updated = 2;
```
amyangfei, line 865: Does `Deleted` mean only `delete` event contains the previous value, and `Updated` mean `put` event (insert or update) contains the previous value?
Maybe we need a mode that both `delete` and `put` contain the previous value.

proto/kvrpcpb.proto
```diff
@@ -858,3 +858,9 @@ message KeyRange {
     bytes start_key = 1;
     bytes end_key = 2;
 }
+
+enum ExtraRead {
```
MyonKeminta, line 862: Why is this defined in `kvrpcpb.proto`?

5kbpers: Both `storage` and `raftstore` will use it.

overvenus: It's okay to put them in cdcpb as cdcpb does not depend on both `raftstore` and `storage`.

proto/kvrpcpb.proto
```diff
@@ -861,3 +861,9 @@ message KeyRange {
     bytes start_key = 1;
     bytes end_key = 2;
 }
+
+enum ExtraOp {
+    Noop = 0;
+    ReadDeleted = 1;
+    ReadOldValue = 2;
```
overvenus, line 868: Can we have them both set in the future?

5kbpers: `ReadDeleted` is included in `ReadOldValue`.

overvenus: Could you add some comments? The behavior is not obvious to me.

proto/kvrpcpb.proto
5kbpers: But it's a switch which could control the operation in `txn`, so I guess it's a little strange to put it to cdcpb.🤔

PR #634, opened 2020-06-23
Add proto for parallel commit

sticnarf:
This PR adds initial supports for parallel commit according to this [design doc](https://github.com/tikv/sig-transaction/blob/master/design/parallel-commit/initial-design.md).
It should suffice for the first iteration.


nrc:
I sent a pr to the design doc to remove the bools since I think we can just use the list of secondary keys. What do you think?

youjiali1995:
> I sent a pr to the design doc to remove the bools since I think we can just use the list of secondary keys. What do you think?

Do we need to know whether it is a parallel commit transaction from a secondary lock?

nrc:
> > I sent a pr to the design doc to remove the bools since I think we can just use the list of secondary keys. What do you think?
> 
> 
> 
> Do we need to know whether it is a parallel commit transaction from a secondary lock? If we don't, I think we can remove the bool.

I guess we do since tikv will need to compute a time stamp

PR #637, opened 2020-06-29
Add async commit fields to LockInfo

MyonKeminta:
This PR added fields about parallel commit to `LockInfo`.


proto/kvrpcpb.proto
```diff
@@ -711,6 +711,10 @@ message LockInfo {
     uint64 txn_size = 5;
     Op lock_type = 6;
     uint64 lock_for_update_ts = 7;
+    // Fields for transactions that are using Parallel Commit.
+    bool use_parallel_commit = 8;
```
youjiali1995, line 715: How about renaming it to async_commit?

sticnarf: Oh, we should rename all occurances of "parallel commit" to "async commit"?

youjiali1995: Yes.

MyonKeminta: Ok. Let me do this in this PR.

MyonKeminta:
@youjiali1995 PTAL again thanks!

PR #642, opened 2020-07-02
br: add compression type for br backup request

glorv:
add `CompressionType` for br BackupRequest, so br can choose compress algorithm explicitly.


proto/backup.proto
```diff
@@ -116,6 +116,15 @@ message BackupRequest {
...
     bool is_raw_kv = 10;
     string cf = 11;
+    
+    // algorithm used for compress sst file
+    enum CompressionType {
+        UNKNOWN = 0;
+        LZ4 = 1;
+        SNAPPY = 2;
+        ZSTD = 3;
+    }
```
overvenus, line 126: Could you define the enum outside of `BackupRequest`?

PR #646, opened 2020-07-10
pdpb: include down/pending peers in scan region response

disksing:


proto/pdpb.proto
```diff
@@ -224,11 +224,25 @@ message ScanRegionsRequest {
     bytes end_key = 4; // end_key is +inf when it is empty.
 }
 
+message Region {
```
rleungx, line 227: It seems we can use it for some other responses like `GetRegionResponse`

disksing: yes. But GetRegion already has pending and down peers fields. And it will need some efforts to keep it compatible.

PR #651, opened 2020-07-17
Add a CheckSecondaryLocks message

nrc:
A command for async commit for checking locks and leaving a rollback tombstone if the lock has not got a success status.

PTAL @sticnarf @cfzjywxk

### DIFF
--- a/proto/kvrpcpb.proto
+++ b/proto/kvrpcpb.proto
@@ -212,6 +212,27 @@ message CheckTxnStatusResponse {
     repeated bytes secondaries = 7;
 }
 
+// Part of the async commit protocol, checks for locks on all supplied keys. If a lock is missing,
+// does not have a successful status, or belongs to another transaction, TiKV will leave a rollback
+// tombstone for that key.
+message CheckSecondaryLocksRequest {
+    Context context = 1;
+    repeated bytes keys = 2;
+    // Identifies the transaction we are investigating.
+    uint64 start_version = 3;
+}
+
+message CheckSecondaryLocksResponse {
+    errorpb.Error region_error = 1;
+    // For each key in `keys` in `CheckSecondaryLocks`, there will be a lock in
+    // this list if there is a lock present and belonging to the correct transaction,
+    // nil otherwise.
+    repeated LockInfo locks = 2;
+    // If any of the locks have been committed, this is the commit ts used. If no
+    // locks have been committed, it will be zero.
+    uint64 commit_ts = 3;
+}
+
 // The second phase of writing to TiKV. If there are no errors or conflicts, then this request
 // commits a transaction so that its data can be read by other transactions.
 message CommitRequest {
@@ -313,7 +334,7 @@ message ScanLockResponse {
 // commit or rollback the transaction and unlock the key.
 message ResolveLockRequest {
     Context context = 1;
-    uint64 start_version  = 2;
+    uint64 start_version = 2;
     // `commit_version == 0` means the transaction was rolled back.
     // `commit_version > 0` means the transaction was committed at the given timestamp.
     uint64 commit_version = 3;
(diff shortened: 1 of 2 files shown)


proto/kvrpcpb.proto
```diff
@@ -212,6 +212,24 @@ message CheckTxnStatusResponse {
...
+    repeated bytes keys = 2;
+    // Identifies the transaction we are investigating.
+    uint64 start_version = 3;
+}
+
+message CheckSecondaryLocksResponse {
+    errorpb.Error region_error = 1;
+    // For each key in `keys` in `CheckSecondaryLocks`, there will be a lock in
+    // this list if there is a lock present and belonging to the correct transaction,
+    // nil otherwise.
```
sticnarf, line 229: If only part of keys exist, I think we should also return some information. For example, we need to know whether we should fallback to the normal way (because some keys are not using async commit). Or if some locks have been committed, we can directly use that information to resolve locks.

nrc: As I understand it, we would only use this request after doing a `CheckTxnStatus` to find the primary, so for the lock we care about, we are guaranteed for it to be using async commit, and for other locks, we don't need to resolve them. Other locks being resolved will have their own messages being sent so I don't think we can helpfully use information in response to this request.

sticnarf: It is possible that the primary key is using async commit. But a secondary key fails to do an async commit for some unknown reason.

nrc: I'm not sure how that fallback works? In my understanding if a secondary lock fails, then we just rollback the transaction.

sticnarf: Currently we are using a PD timestamp. So if we fail to get a PD timestamp, we can fallback. (Though we don't do it know) 
In the future, if we are using a calculated timestamp, we can't derive a timestamp from a calculated timestamp. Then, we have to fallback to the normal mode.

nrc: Thanks, so we can fallback at the prewrite stage, but not here in the resolve phase?

Since TTL must have expired before sending this message and we will only send this if the primary lock has not been committed, then I think we must always rollback if it is a fallback transaction (or if there is a failed or missing lock). If all we are doing is a rollback for the transaction, then I think we don't need any more information than we have already. Is something different?

cfzjywxk: The fallback logic is [here](https://github.com/pingcap/tidb/pull/18622/files#diff-e02e820b9bc11d6e476f1e6641f41c6fR136) in the prewrite stage.  If the transaction is still using async commit protocol and some locks are not found, it's possible they are committed by another concurrent recovery process, in this case, we need to return the already commit message to the caller like the `commitTS`.

cfzjywxk:
If TIDB crashes after prewrite finished but before some valid checks like schema check, `commitTS`  check. The async commit transaction status is uncertain, these checks should be done in the recovery logic.

proto/kvrpcpb.proto
```diff
@@ -212,6 +212,24 @@ message CheckTxnStatusResponse {
...
+// does not have a successful status, or belongs to another transaction, TiKV will leave a rollback
+// tombstone for that key.
+message CheckSecondaryLocksRequest {
+    Context context = 1;
+    repeated bytes keys = 2;
+    // Identifies the transaction we are investigating.
+    uint64 start_version = 3;
+}
+
+message CheckSecondaryLocksResponse {
```
youjiali1995, line 225: Need a `commit_version`.

nrc:
Rebased and comment addressed. PTAL @sticnarf @youjiali1995

sticnarf:
@youjiali1995 I think we still need to seek the write.
If we would write a rollback as long as the lock does not exist, we might write rollback of a key that is committed...
And getting the commit ts can potentially save us an RPC.

youjiali1995:
> If we would write a rollback as long as the lock does not exist, we might write rollback of a key that is committed...

It's not true. If the secondary lock doesn't exist, we must use `check_txn_status` to roll back the primary lock or get the status of it at first.

PR #652, opened 2020-07-17
add CheckAsyncCommitLocks RPC

sticnarf:
As mentioned in https://github.com/pingcap/tidb/pull/18467#discussion_r455613079, we need a command that checks and roll back secondary locks of an async commit transaction atomically. So I add this RPC proto.

This RPC should be used only for transactions with secondary locks. (We shouldn't call it for a single key transaction.)

If `commit_ts > 0`, this transaction has been committed. We can use that information to resolve locks.

If the returned `use_async_commit` is false, we should fallback to the normal way of resolving lock.

If `use_async_commit` is true but `min_commit_ts` is 0, this means some key does not exist and we have written a rollback record for it. We can roll back this transaction.


sticnarf:
Oh #651 is for the same thing.

PR #657, opened 2020-07-26
Add a key error field to CheckSecondaryLocks

nrc:
Was missing from the initial PR.

PTAL @sticnarf @youjiali1995


sticnarf:
I think we can add it. But I don't know when we will get a KeyError from this command. I thought in no case would we get a KeyError from this command...

youjiali1995:
> I think we can add it. But I don't know when we will get a KeyError from this command. I thought in no case would we get a KeyError from this command...

So, if some locks don't exist, the response contains an shorter `LockInfo` vector?

sticnarf:
> > I think we can add it. But I don't know when we will get a KeyError from this command. I thought in no case would we get a KeyError from this command...
> 
> So, if some locks don't exist, the response contains an shorter `LockInfo` vector?

If some locks don't exist, the lock info is just empty.

youjiali1995:
> > > I think we can add it. But I don't know when we will get a KeyError from this command. I thought in no case would we get a KeyError from this command...
> > 
> > 
> > So, if some locks don't exist, the response contains an shorter `LockInfo` vector?
> 
> If some locks don't exist, the lock info is just empty.

Make sense. And I think the `KeyError` is not necessary...

nrc:
Having written the actual code, I also don't think this is necessary. Closing for now, sorry for the noise.

sticnarf:
@youjiali1995 @nrc 
The key error can occur in bad format in the lock and write CF.
Although I think it's OK to panic in this case because there must be some data corruption, I prefer to return key error to the client to keep consistent with other commands.

PR #658, opened 2020-07-31
Support joint consensus

NingLin-P:


NingLin-P:
PTAL @BusyJay @nolouch

proto/raft_cmdpb.proto
```diff
@@ -215,6 +219,7 @@ message AdminRequest {
     CommitMergeRequest commit_merge = 8;
     RollbackMergeRequest rollback_merge = 9;
     BatchSplitRequest splits = 10;
+    ChangePeerV2Request change_peer_v2 = 11;
```
BusyJay, line 222: I suggest to add a new `AdminCmdType` to avoid potential compatibility bug.

proto/raft_cmdpb.proto
```diff
@@ -215,6 +220,7 @@ message AdminRequest {
...
```
BusyJay, line 223: Should add corresponding response too.

proto/raft_cmdpb.proto
```diff
@@ -116,6 +116,14 @@ message ChangePeerResponse {
     metapb.Region region = 1;
 }
 
+message ChangePeerV2Request {
+    repeated ChangePeerRequest changes = 1;
+}
+
+message ChangePeerV2Response {
```
BusyJay, line 123: It's not used.

NingLin-P: Fixed.

PR #659, opened 2020-08-03
pdpb: revert "pdpb: include down/pending peers in scan region response (#646)"

nrc:
This reverts #646 since it breaks backwards compatibility and prevents updating TiDB:

```
➜  tidb git:(master) ✗ make
CGO_ENABLED=1 GO111MODULE=on go build  -tags codes  -ldflags '-X "github.com/pingcap/parser/mysql.TiDBReleaseVersion=v4.0.0-beta.2-872-g03003538d-dirty" -X "github.com/pingcap/tidb/util/versioninfo.TiDBBuildTS=2020-08-03 03:25:04" -X "github.com/pingcap/tidb/util/versioninfo.TiDBGitHash=03003538d1d7a1762e02fc62365e4832f9df4dd7" -X "github.com/pingcap/tidb/util/versioninfo.TiDBGitBranch=master" -X "github.com/pingcap/tidb/util/versioninfo.TiDBEdition=Community" ' -o bin/tidb-server tidb-server/main.go
# github.com/pingcap/pd/v4/client
../../gopath/pkg/mod/github.com/pingcap/pd/v4@v4.0.0-rc.2.0.20200714122454-1a64f969cb3c/client/client.go:562:24: cannot use resp.GetRegions() (type []*pdpb.Region) as type []*metapb.Region in return argument
make: *** [server] Error 2
```

PTAL @disksing @rleungx


disksing:
It seems we should be able to upgrade by migrating some function calls to the new interface. Let me take a look.

nrc:
The TiDB build issue is addressed in https://github.com/pingcap/tidb/pull/18938, aiui, the back compat issue will not be a problem at runtime because of the `region_metas` field and kvproto is not a published crate, so I think we can close this for now.

PR #663, opened 2020-08-17
cdc: add regions field for reducing message of resolved ts

5kbpers:

Add `regions` field into `cdcpb::Event`.

Combine resolved ts messages for multiple regions into one.


proto/cdcpb.proto
```diff
@@ -85,6 +85,7 @@ message Event {
     uint64 region_id = 1;
     uint64 index = 2;
     uint64 request_id = 7;
+    repeated uint64 regions = 9;
```
overvenus, line 88: Could you deprecate `event.resolved_ts` and add a new field in `ChangeDataEvent` at line 102?

PR #586, opened 2020-03-25
support batch coprocessor

hanfei1991:

PR #664, opened 2020-08-17
[4.0] Pick serval PRs for ticdc compiling

5kbpers:
pick #584 #594 #595 #597 #586 #603 #581


proto/cdcpb.proto
```diff
@@ -69,6 +69,7 @@ message Event {
     uint64 region_id = 1;
     uint64 index = 2;
     uint64 request_id = 7;
+    repeated uint64 regions = 8;
```
overvenus, line 72: Could you revert the change? So we can merge other cherry-picks first.

PR #665, opened 2020-08-20
support checking memory locks at read index

sticnarf:
With async commit, locks should be visible to readers before they are written to the raftstore. This is also true for replica reads.

The unapplied locks are stored in the memory of the leader. To make followers or learners able to see the unapplied locks, necessary information should be sent to the leader in `ReadIndex` to check if any lock in memory.


5kbpers (approved):

youjiali1995 (approved):

proto/kvrpcpb.proto
```diff
@@ -561,11 +561,19 @@ message SplitRegionResponse {
 // Sent from TiFlash to a TiKV node.
 message ReadIndexRequest{
     Context context = 1;
+
+    // TiKV checks the given range if there is any unapplied lock
+    // blocking the read request.
+    uint64 start_ts = 2;
+    repeated KeyRange ranges = 3;
```
solotzg, line 568: What if `ranges` is empty ? Does it mean full range of this region ?

sticnarf: If `ranges` is empty, no range will be checked. The start and end bound of the region should be passed if ithe full range needs to be checked.

MyonKeminta:
Is there any other problem with this PR? If no I think it can be merged

sticnarf:
> Is there any other problem with this PR? If no I think it can be merged

I think https://github.com/tikv/tikv/pull/8669 should be reviewed first. If there's no problem for that PR then we can merge it.

PR #666, opened 2020-08-20
Add encryption meta to metapd.Region

yiwu-arbug:
Adding encryption meta to `metapb.Region`. This is going to be used by PD to encrypt the start_key and end_key when persisting the `Region` struct.

Signed-off-by: Yi Wu <yiwu@pingcap.com>


proto/encryptionpb.proto
```diff
@@ -11,6 +11,16 @@ option (rustproto.lite_runtime_all) = true;
 
 option java_package = "org.tikv.kvproto";
 
+// General encryption metadata for any data type.
+message EncryptionMeta {
+    // Whether data is encrypted.
+    bool is_encrypted = 1;
```
HunDunDM, line 17: `EncryptionMeta` is not `nil`, but `is_encrypted` is `false`. What is the corresponding situation?

HunDunDM (commented):
rest LGTM

proto/encryptionpb.proto
yiwu-arbug: The two case should both treated as the region boundary keys are not encrypted.

yiwu-arbug:
updated. PTAL

HunDunDM (approved):
LGTM

Yisaer (approved):
LGTM

PR #667, opened 2020-08-21
[4.0] cdc: add regions field for reducing message of resolved ts

5kbpers:

pick #663


5kbpers:
> why `uint64 resolved_ts = 6 [deprecated=true]` is not picked

My fault, already fixed.

PR #668, opened 2020-08-24
add MaxTimestampNotSynced region error

sticnarf:
If a peer has just become leader but the timestamp is not synced to the concurrency manager, prewrite operation can return a `MaxTimestampNotSynced` error to client. Then, the client can backoff and retry.


cfzjywxk:
LGTM

youjiali1995 (commented):
How about returning 0(`min_commit_ts`) and falling back to normal 2PC?

sticnarf:
> How about returning 0(`min_commit_ts`) and falling back to normal 2PC?

I don't have perference and it was my original idea.

But sometime before it seems we remove the possibility of fallbacks in a few PRs (like force checking `use_async_commit`)... Should we change them back?

youjiali1995:
> > How about returning 0(`min_commit_ts`) and falling back to normal 2PC?
> 
> I don't have perference and it was my original idea.
> 
> But sometime before it seems we remove the possibility of fallbacks in a few PRs (like force checking `use_async_commit`)... Should we change them back?

I think we must support fallback considering it's an RPC call to get a ts from PD. Alternatively, TiDB can back off a little and switch to normal 2PC, but I think it's harder to implement it than in TiKV.

MyonKeminta:
I think the easiest way is that once a transaction tries to commit in async commit mode, it will never fallback to 2PC, which is what this PR is going to do. I think falling back can be an optimization that we can do later, and I'm afraid it might not be as easy as removing some checks.

MyonKeminta:
Actually I think it would be better if it's possible to wait for TS being ready in tikv directly.

sticnarf:
> Actually I think it would be better if it's possible to wait for TS being ready in tikv directly.

Of course. But it's technically a bit difficult with the current TiKV implementation (I don't have a good idea how to implement it easily)

sticnarf:
> I think the easiest way is that once a transaction tries to commit in async commit mode, it will never fallback to 2PC, which is what this PR is going to do. I think falling back can be an optimization that we can do later, and I'm afraid it might not be as easy as removing some checks.

I agree that it's easier to be correct if we do not support fallback. This protocol does not rule out the possibility of fallback. I think we can use this first and take fallback as an optimization. What do you think? @youjiali1995

youjiali1995:
> > I think the easiest way is that once a transaction tries to commit in async commit mode, it will never fallback to 2PC, which is what this PR is going to do. I think falling back can be an optimization that we can do later, and I'm afraid it might not be as easy as removing some checks.
> 
> I agree that it's easier to be correct if we do not support fallback. This protocol does not rule out the possibility of fallback. I think we can use this first and take fallback as an optimization. What do you think? @youjiali1995

OK

youjiali1995 (approved):

MyonKeminta (approved):

nrc (approved):

PR #671, opened 2020-08-27
add lock info to CheckTxnStatusResponse

sticnarf:
For async commit recovery, we need the `min_commit_ts` of all locks. We can get secondary locks via `CheckSecondaryLocks` but we cannot get the `min_commit_ts` of the primary lock now.

We should carry the information of the whole primary lock in the response of `CheckTxnStatus`.


proto/kvrpcpb.proto
```diff
@@ -210,6 +210,7 @@ message CheckTxnStatusResponse {
     Action action = 5;
     bool use_async_commit = 6;
     repeated bytes secondaries = 7;
+    uint64 min_commit_ts = 8;
```
youjiali1995, line 213: How about returning `LockInfo` instead?

sticnarf: sounds better

sticnarf:
Note this is an incomipatible change (but is master-to-master). Do not merge it until the corresponding changes are made and reviewed in the related projects (TiKV and unistore).

coocood:
@sticnarf @youjiali1995 
We already have a KeyError field and that has a LockInfo, can we reuse it?

sticnarf:
@coocood But it will be semantically strange because it's not an error for `CheckTxnStatus`...

coocood:
LGTM

proto/kvrpcpb.proto
```diff
@@ -198,6 +198,7 @@ message CheckTxnStatusRequest {
 }
 
 message CheckTxnStatusResponse {
+    reserved 6, 7;
```
nrc, line 201: Why do we need to reserve 6 and 7, this proto has never been used in a published version, only on master, so it should be OK to make a breaking change.

sticnarf: Reserving these fields helps us pass the integration tests.
If we just change the type of the field, when we first upgrade TiKV, the old version of TiDB will receive a proto it cannot handle. And vice versa. Then, we have to force merge a TiKV PR or TiDB PR.

nrc: Can we remove them later? Seems better to force merge once, rather than live with cruft forever

sticnarf: I didn't think it's cruft... Will reserved fields have extra cost? If so, we can force merge at first.

nrc: I don't think it there is performance cost, but it is just a bit ugly to read. It is ok now, but if every proto has something like `reserved 6, 7, 10, 11, 12, 15, 16, 17;` or something

youjiali1995: Agree, it's ugly.

sticnarf: ok.

sticnarf:
@coocood @nrc @youjiali1995 
Please also take a look at https://github.com/tikv/tikv/pull/8529 and https://github.com/ngaut/unistore/pull/418
If they look good, we can merge this PR and quickly merge  and TiDB. (to make broken time as little as possible)

pkg/raft_cmdpb/raft_cmdpb.pb.go
```diff
@@ -71,7 +71,7 @@ func (x CmdType) String() string {
 	return proto.EnumName(CmdType_name, int32(x))
 }
 func (CmdType) EnumDescriptor() ([]byte, []int) {
-	return fileDescriptor_raft_cmdpb_5dc25bc5f43956ae, []int{0}
+	return fileDescriptor_raft_cmdpb_8fa9c3cdf0171bc7, []int{0}
```
youjiali1995, line 74: Why this file is changed?

sticnarf: I don't know. I cleaned and regenerate, but this file is also changed.

sticnarf: I use protoc 3.8.0 following the README. Maybe the last commit didn't use the expected version?

youjiali1995 (approved):

pkg/raft_cmdpb/raft_cmdpb.pb.go
sticnarf: I think it's because https://github.com/pingcap/kvproto/pull/644 didn't generate the go proto.

MyonKeminta (approved):

proto/kvrpcpb.proto
```diff
@@ -208,8 +208,7 @@ message CheckTxnStatusResponse {
     uint64 commit_version = 4;
     // The action performed by TiKV (and why if the action is to rollback).
     Action action = 5;
-    bool use_async_commit = 6;
-    repeated bytes secondaries = 7;
+    LockInfo lock_info = 6;
```
overvenus, line 211: Reusing the same tag is a breaking change.

proto/kvrpcpb.proto
overvenus: It's indeed a breaking change and forces BR to upgrade kvproto and TiDB https://github.com/pingcap/br/pull/481.

I suggest before we make any change to kvproto we should first make sure the feature development and test are almost complete.

nrc: > I suggest before we make any change to kvproto we should first make sure the feature development and test are almost complete.

It seems like we should have some conventions, certianly

proto/kvrpcpb.proto
MyonKeminta: It's ok since it's only on master. But I'm afraid if other projects need to struggle to get them compiled.

PR #674, opened 2020-09-08
cdc: add ticdc version to request header

overvenus:
Add ticdc version to request header, so that TiKV knows what's the version of a downstream ticdc.


amyangfei (approved):

amyangfei (commented):
Should we add another `ExtraOp` type, such as `EnableBatchResolvedTs` to control whether enabling this feature, just as what we do with the `old value` feature

MyonKeminta (approved):

overvenus:
I'm not sure batch resolved ts fits in `kvrpc.ExtraOp` (we should not leak cdc to other packages). Batch resolved ts should be enabled for all changefeed when it's stable (so maybe it does not need a flag?). Anyway, this PR adds version to request, we can add batch resolved ts flag in later PR if it's needed.

PR #676, opened 2020-09-08
sst_importer: add delete type for write interface

3pointer:
add delete type for each kv pair in `sst_importer.Write`. for PiTR and Lightning


proto/import_sstpb.proto
```diff
@@ -185,6 +185,11 @@ message SetDownloadSpeedLimitResponse {
 message Pair {
     bytes key = 1;
     bytes value = 2;
+    enum OP {
+        Put = 0;
+        Delete = 1;
```
overvenus, line 190: Tag 0 in enum is the default value. It's better to reserve the tag for Unknown to avoid surprising behavior.
```suggestion
        Unknown = 0;
        Put = 1;
        Delete = 2;
```

3pointer: I think Put = 0 is better, because lightning local backend already use this protocol. we can avoid incompatible during upgrade.

overvenus (approved):

glorv:
LGTM

glorv (approved):

PR #678, opened 2020-09-15
Add SyncMaxTS method

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>

Part of tikv/pd#2759.


Yisaer (commented):
Basically LGTM. I'm wondering maybe we also need to add the sender information like memberID in request. And each receiver shoudl compare the sender member ID and the leaderid in their vision.

Yisaer (approved):
LGTM @disksing  PTAL

disksing (approved):

proto/pdpb.proto
```diff
@@ -627,4 +627,5 @@ message SyncMaxTSRequest {
 message SyncMaxTSResponse {
     ResponseHeader header = 1;
     Timestamp max_local_ts = 2;
+    repeated string synced_dcs = 3;
```
disksing, line 630: Maybe just `dcs`, for the first phase they do not sync anything.

JmPotato: nice idea.

disksing (approved):

PR #680, opened 2020-09-27
Add single-region 1PC protocol

MyonKeminta:
This PR adds protocol for supporting 1PC for transactions that involves only one single region.


MyonKeminta:
I'm still thinking if there are other corner cases that we didn't notice before.

coocood:
LGTM

sticnarf:
LGTM

sticnarf (approved):

youjiali1995 (approved):

PR #684, opened 2020-10-19
allow async commit to use max_commit_ts limit

sticnarf:
This PR renames `one_pc_max_commit_ts` to `max_commit_ts` as we want to also use this field limit async-commit prewrites.

Async-commit prewrites don't report error by the `one_pc_commit_ts` field, so this PR also adds an error message for `max_commit_ts` exceeded.


coocood:
LGTM

MyonKeminta (approved):
LGTM

youjiali1995 (approved):

PR #689, opened 2020-11-10
Improve ScanDetail and ExecDetails

breeswish:
1. Improvement: Supply many comments for ScanDetailV2 and TimeDetail fields.

2. Improvement: Rename message `HandleTime` to `TimeDetail`, field `handle_time` to `time_detail`, `bool handle_time` to `bool record_time_stat`, etc.

   Field and message rename is upgrade compatible. However code need to be updated accordingly.

3. Add: In GetResponse and BatchGetResponse, use `ExecDetailsV2` instead of `ScanDetailV2`, so that time can be reported as well in future, as proposed in https://github.com/tikv/tikv/issues/8942.


proto/kvrpcpb.proto
```diff
@@ -817,39 +821,76 @@ enum IsolationLevel {
...
-    int64 wait_ms = 1;
-    // Processing time, excluding wait time.
-    int64 process_ms = 2;
+message TimeDetail {
+    // Off-cpu wall time elapsed in TiKV side. Usually this includes queue waiting time and
+    // other kind of waitings in series.
+    int64 wait_wall_time_ms = 1;
+    // Off-cpu and on-cpu wall time elapsed to actually process the request payload. It does not
+    // include `wait_wall_time`.
+    // This field is very close to the CPU time in most cases. Some wait time spend in RocksDB
```
zhongzc, line 830: Do we really have the ability to measure the CPU time of a single request, even we don't take Mutex wait time into account?

breeswish: The precise CPU time can be measured by user_cpu_time + system_cpu_time. Not sure about the cost though.

zhongzc: IMO, this field is not close to CPU time. It's a typical wall time.

breeswish: Oh yes. I only want to express that, in 99% cases, the higher this field is, the higher CPU usage there will be.

zhongzc (approved):

proto/kvrpcpb.proto
```diff
@@ -817,39 +821,76 @@ enum IsolationLevel {
...
 message ExecDetails {
-    HandleTime handle_time = 1;// set when ctx.handle_time = true or meet slow query
-    ScanDetail scan_detail = 2;// set when ctx.scan_detail = true or meet slow query
-    bool use_scan_detail_v2 = 3;// set when ctx.scan_detail = true or meet slow query
-    ScanDetailV2 scan_detail_v2 = 4;// set when ctx.scan_detail = true or meet slow query
+    // Available when ctx.record_time_stat = true or meet slow query.
+    TimeDetail time_detail = 1;
+
+    // Available when ctx.record_scan_stat = true or meet slow query.
+    ScanDetail scan_detail = 2;
```
zhongzc, line 885: Maybe `reserved`s are needed here.

breeswish: Good catch

breeswish:
@crazycs520 PTAL if you are free, thanks!

crazycs520 (approved):
LGTM

proto/kvrpcpb.proto
```diff
@@ -36,8 +36,9 @@ message GetResponse {
     bytes value = 3;
     // True if the key does not exist in the database.
     bool not_found = 4;
-    // Set if ctx.scan_detail = true or meet slow query.
-    ScanDetailV2 scan_detail_v2 = 5;
+    reserved 5;
```
sticnarf, line 39: This field is not used in a released version. Do we really need to reserve it?

breeswish: Only for our lovely developers who are using the master branch.

PR #700, opened 2020-12-07
Allow CheckTxnStatus roll back async-commit locks

sticnarf:
Design: https://github.com/tikv/sig-transaction/issues/64#issuecomment-738530608


proto/kvrpcpb.proto
```diff
@@ -212,6 +212,9 @@ message CheckTxnStatusRequest {
     // If true, then TiKV will leave a rollback tombstone in the write CF for `primary_key`, even if
     // that key is not locked.
     bool rollback_if_not_exist = 6;
+    // This field is set to true only if the transaction is known to fall back from async commit.
+    // CheckTxnStatus treats the transaction as non-async-commit if this field is true.
+    bool async_commit_fallback = 7;
```
nrc, line 217: I think it would be better to name the field for the effect it has, rather than what caused it, so that if it is used for something else in the future then the name is not confusing. Maybe `force_sync_commit`?

sticnarf: OK

nrc (approved):
lgtm

youjiali1995 (approved):

PR #703, opened 2020-12-10
*: support stale read

NingLin-P:


proto/kvrpcpb.proto
```diff
@@ -769,6 +769,10 @@ message Context {
     // A hint for TiKV to schedule tasks more fairly. Query with same task ID
     // may share same priority and resource quota.
     uint64 task_id = 16;
+
+    // No required to read the most up-to-date data, replicas with `safe_ts` >= `start_ts`
```
BusyJay, line 773:
```suggestion
    // Not required to read the most up-to-date data, replicas with `safe_ts` >= `start_ts`
```

proto/raft_serverpb.proto
```diff
@@ -23,6 +23,9 @@ message RaftMessage {
     // If it has value, to_peer should be removed if merge is never going to complete.
     metapb.Region merge_target = 9;
     ExtraMessage extra_msg = 10;
+    // use for stale read
+    uint64 applied_index = 11;
```
BusyJay, line 27: Why is it necessary?

proto/raft_serverpb.proto
```diff
@@ -23,6 +23,9 @@ message RaftMessage {
...
+    uint64 safe_ts = 12;
```
BusyJay, line 28: I don't think this is something raft store can understand.

proto/raft_serverpb.proto
NingLin-P: To forward follower's `safe_ts` through raft heartbeat

BusyJay (approved):

BusyJay (approved):

nolouch (approved):
lgtm

PR #704, opened 2020-12-14
Add 1PC flag to raft command request

sticnarf:
CDC needs to handle 1PC transactions in special manners. To let CDC know which mutations are from 1PC transactions, this flag is added.

Actually, this flag does not need a raft replication. But if we put it inside a raft request, we needn't modify raftstore interfaces to send it to CDC, which helps us avoid complex code changes.


sticnarf:
@BusyJay Could you please take a look whether this is a good design?

BusyJay:
I suggest to add the flag to Headers of RaftCmdRequest. And OnePC is a flag from transaction layer, so it should probably not show up in the raft protocol.

sticnarf:
> I suggest to add the flag to Headers of RaftCmdRequest. And OnePC is a flag from transaction layer, so it should probably not show up in the raft protocol.

OK. I will move the flags to the header.

For the flag bit definition, right, it's strange to have `OnePC` in raft protocol. But there should be a place to guarantee the uniqueness of each bit. So maybe we have to put it there...

BusyJay:
For example, you can put it in kvrpcpb.proto. The definition of flags and bits don't have to be put in one place. Flags are more like a customized context, it can be arbitrary value that coprocessor can understand.

nrc (approved):
lgtm

proto/kvrpcpb.proto
```diff
@@ -1009,3 +1009,11 @@ message CheckLeaderResponse {
     repeated uint64 regions = 1;
     uint64 ts = 2;
 }
+
+// Additional flags for a write batch. They should be set in the `flags` field in `RaftRequestHeader`.
+enum WriteBatchFlag {
```
5kbpers, line 1014: Not necessary to define these flags in proto.

sticnarf: OK. I decide to put it directly inside TiKV.

5kbpers (approved):
LGTM

MyonKeminta (approved):
LGTM

PR #705, opened 2020-12-15
txn: update the kvproto for check txn status api

cfzjywxk:
Make the check result certain so that the resolved transactions are saved correctly in resolved cache.

Add a field `resolving_pessimistic_lock` to indicate if the check request is used to resolve a pessimitic lock.
Add action types `TTLExpirePessimisticRollback` and `LockNotExistDoNothing` so that calller could know if the transaction status is uncertain, cache the resolved results only if the status is derterministic.

Update the protocol for the two pathes in which the transaction status may be uncertain and is still ongoing:
```
if the primary key lock exists
    if resolving_pessimistic_lock and primary key lock is pessimistic lock
         if pk lock is expired
            just pessimistic rollback pk lock and set action to `TTLExpirePessimisticRollback`
else
    if resolving_pessimistic_lock  and primary key lock is not found and there is no commit or rollback record
        set action to `LockNotExistDoNothing` and do not write a rollback record
```

For cache resolved transaction:
```
if ttl == 0
    if comitted
        cache it
    else
        cache if for action type `NoAction`, `TtlExpireRollback`, `LockNotExistRollback`
```


nrc (approved):
lgtm

youjiali1995 (approved):

sticnarf:
LGTM

PR #707, opened 2021-01-05
Add for_balance for snapshot meta

Connor1996:
Add `for_balance` for snapshot meta. It's used to indicate whether the snapshot is triggered by load-balance.


yiwu-arbug (approved):
LGTM

hicqu (approved):

PR #714, opened 2021-01-22
Add endkey field to ScanLock request

MyonKeminta:
This PR adds endKey field to ScanLockRequest. To fix / https://github.com/tikv/tikv/issues/9526 for non-green-gc mode, TiKV needs to update max_ts, then check memory locks for the range, and finally get a snapshot for scanning. The endKey field is added to limit the range for checking memory lock.


sticnarf (approved):
LGTM

cfzjywxk:
LGTM

youjiali1995 (approved):

PR #717, opened 2021-02-01
backup: add br/kv version info in backup meta

lichunzhu:
Add `br_version` and `tikv_versions` in `backup.proto` to save version info for debug.


proto/backup.proto
```diff
@@ -43,6 +43,11 @@ message BackupMeta {
 
     // In incremental backup, DDLs which are completed in (lastBackupTS, backupTS] will be stored here.
     bytes ddls = 10;
+
+    // Save the br commit to run backup jobs
+    string br_version = 11;
+    // A map of the tikv store ID to store version
+    map<uint64, string> tikv_versions = 12;
```
glorv, line 50: I'm a bit confused with this map. If some of the tikv instances' version are different, what can we do with them?

lichunzhu: It will be simpler for br to debug. If we fail to ingest an SST file because of the version compatibility problem, we can get  store_id from this file's name and version from this map.

overvenus, line 50: I don't think we need this, it should be the same as cluster version.

proto/backup.proto
```diff
@@ -43,6 +43,11 @@ message BackupMeta {
 
     // In incremental backup, DDLs which are completed in (lastBackupTS, backupTS] will be stored here.
     bytes ddls = 10;
+
+    // Save the br commit to run backup jobs
```
overvenus, line 47:
```suggestion
    // Save the version of BR running backup jobs.
```

proto/backup.proto
overvenus: Ingest SST happens in restore, but backupmeta is created in backup.

lichunzhu: What if different TiKVs have different versions? By the way, what should I fill in `cluster version`?

lichunzhu: When we fail to ingest sst in restore, we can check the `backupmeta` to find out the tikv's version who backuped this sst file.

overvenus: > What if different TiKVs have different versions?

It's abnormal, or a cluster is in rolling upgrade.

> By the way, what should I fill in cluster version?

https://docs.pingcap.com/tidb/stable/pd-control#:~:text=cluster-version (the version is reported by TiKV)

overvenus: TiKV rarely changes SST file format (still in SST v2). I think it's fine to not record tikv version.

lichunzhu: tikv_versions deleted in https://github.com/pingcap/kvproto/pull/717/commits/9ee61335654225c1fad27beac7806ef389212d64

kennytm (commented):
lgtm

proto/backup.proto
```diff
@@ -43,6 +43,9 @@ message BackupMeta {
 
     // In incremental backup, DDLs which are completed in (lastBackupTS, backupTS] will be stored here.
     bytes ddls = 10;
+
+    // Save the version of BR running backup jobs.
+    string br_version = 11;
```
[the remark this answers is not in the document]
lichunzhu: We already had `cluster_version` in BackupMeta. The index is 2

glorv:
lgtm

glorv (approved):

kennytm (approved):

PR #719, opened 2021-02-05
harden CMakeLists.txt to be self-contained

TisonKun:


cpp/CMakeLists.txt
```diff
@@ -13,9 +15,14 @@ else ()
...
-list(APPEND generated_files ${__tmp})
+file(GLOB __kvproto_sources RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} kvproto/*.cc)
+file(GLOB __kvproto_headers RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} kvproto/*.h)
+
+add_library(kvproto ${__kvproto_sources} ${__kvproto_headers})
+target_include_directories(kvproto PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+target_include_directories(kvproto PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/kvproto)
 
-add_library(kvproto ${generated_files})
-include_directories(${CMAKE_CURRENT_LIST_DIR}/kvproto)
```
TisonKun, line 20: replace `include_directories` with `target_include_directories`

cpp/CMakeLists.txt
```diff
@@ -13,9 +15,14 @@ else ()
...
-target_include_directories(kvproto PUBLIC ./)
```
TisonKun, line 21: replace bare relative path with CMake variable

cpp/CMakeLists.txt
```diff
@@ -13,9 +15,14 @@ else ()
...
+if (NOT PROTOBUF_INCLUDE_DIRS)
+    find_package(Protobuf REQUIRED)
+endif()
+target_include_directories(kvproto PUBLIC ${PROTOBUF_INCLUDE_DIRS})
```
TisonKun, line 28: kvproto itself requires Protobuf, so far it works well because tics add a global `include_directories(${PROTOBUF_INCLUDE_DIRS})`

TisonKun:
@BusyJay @hanfei1991 PTAL. 

New C++ Test takes about 20~30 mins for preparing gRPC environment, do you know any existing config for that?

TisonKun:
> @BusyJay @hanfei1991 PTAL.
> 
> New C++ Test takes about 20~30 mins for preparing gRPC environment, do you know any existing config for that?

for `protoc` binary, we can do the same way as Go Test, but I'm unsure what about platform & version specific gRPC library.

BusyJay:
Can you use cache provided by CI?

TisonKun:
> Can you use cache provided by CI?

Thanks for your advice. Now the test time has been reduced to 5 mins.

I think the patch is ready to merge. PTAL @BusyJay

TisonKun:
Let the C++ test be required seems a maintainer action, please take care of it.

TisonKun:
@BusyJay @hanfei1991 PTAL

BusyJay (approved):
LGTM except cmake changes. @hanfei1991 may take a look.

TisonKun:
@hanfei1991 PTAL I've tested with tics build.

hanfei1991 (approved):
LGTM

hanfei1991:
Nice work!

TisonKun:
Thanks for your review!

PR #720, opened 2021-02-07
update grpcio to make kvproto build on apple silicon

sticnarf:
Wait for:
- [x] https://github.com/tikv/grpc-rs/pull/507


Cargo.toml
```diff
@@ -16,17 +16,17 @@ prost-codec = ["prost", "prost-derive", "lazy_static", "protobuf-build/grpcio-pr
 
 [dependencies]
 protobuf = "=2.8.0"
-prost = { version = "0.6", optional = true }
-prost-derive = { version = "0.6", optional = true }
+prost = { version = "0.7", optional = true }
+prost-derive = { version = "0.7", optional = true }
 futures = "0.3.5"
-grpcio = { version = "0.7.0", default-features = false, features = ["secure"] }
+grpcio = { version = "0.8.0", default-features = false, features = ["secure", "use-bindgen"] }
```
BusyJay, line 22: I suggest to use following *long* target filter:
```
+[target.'cfg(any(not(linux), not(any(target_arch = "x86_64", target_arch = "aarch64"))))'.dependencies.grpcio]
+version = "0.8.0"
+default-features = false
+features = ["secure", "use-bindgen"]
```

So bindgen will not be required again for all prebuilt releases on linux.

The configuration only works with recent enough rustc.

BusyJay (approved):

Yisaer (approved):

PR #723, opened 2021-02-19
Add keyspace

andylokandy:

### DIFF
--- a/pkg/resource_usage_agent/resource_usage_agent.pb.go
+++ b/pkg/resource_usage_agent/resource_usage_agent.pb.go
@@ -28,7 +28,7 @@ var _ = math.Inf
 // proto package needs to be updated.
 const _ = proto.ProtoPackageIsVersion2 // please upgrade the proto package
 
-type ReportCpuTimeRequest struct {
+type CPUTimeRecord struct {
 	ResourceGroupTag []byte `protobuf:"bytes,1,opt,name=resource_group_tag,json=resourceGroupTag,proto3" json:"resource_group_tag,omitempty"`
 	// UNIX timestamp in second.
 	RecordListTimestampSec []uint64 `protobuf:"varint,2,rep,packed,name=record_list_timestamp_sec,json=recordListTimestampSec" json:"record_list_timestamp_sec,omitempty"`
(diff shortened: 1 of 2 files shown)


proto/pdpb.proto
```diff
@@ -668,3 +679,36 @@ message GetDCLocationInfoResponse {
...
     Timestamp max_ts = 3;
 }
+
+message GetAllKeyspacesRequest {
+    RequestHeader header = 1;
+}
+
+message GetAllKeyspacesResponse {
+    ResponseHeader header = 1;
+    uint64 keyspace_version = 2;
```
sunxiaoguang, line 689: Is it better to make version keyspace specific?

andylokandy: what do you mean?

sunxiaoguang: Looks like the version is global. And if that's true whenever it gets updated all key spaces meta data have to be reloaded.

andylokandy: keyspace meta should be relative small. So I think it won't be a problem. And the global version helps simplify the synchronization process.

proto/pdpb.proto
```diff
@@ -668,3 +679,36 @@ message GetDCLocationInfoResponse {
...
+    // All registered keyspaces.
+    repeated keyspace.Keyspace keyspaces = 3;
+    // Prefixes that were registered to a keyspace and then unbound from it.
+    repeated keyspace.KeyspacePrefix unbound_prefixes = 4;
```
Connor1996, line 693: why need this?

andylokandy: Because `KeyspacePrefix` can't be removed. So it'll leave history even if no keyspace refers to it. With `unbound_prefixes` the users get to know these kind of prefixes and then re-bound it to a keyspace.

Connor1996 (approved):

proto/keyspace.proto
```diff
@@ -0,0 +1,21 @@
...
+enum KeyspaceMode {
+    // RawKv mode.
+    Raw = 0;
+    // RawKv mode with TTL.
+    RawTTL = 1;
+    // Transacation mode.
+    Transaction = 2;
+}
+
+message KeyspacePrefix {
```
BusyJay, line 13: Why make it a separate message?

proto/pdpb.proto
```diff
@@ -668,3 +679,36 @@ message GetDCLocationInfoResponse {
...
+}
+
+message PutKeyspaceRequest {
+    RequestHeader header = 1;
+    keyspace.Keyspace keyspace = 2;
+}
+
+message PutKeyspaceResponse {
+    ResponseHeader header = 1;
+    uint64 keyspace_version = 2;
```
BusyJay, line 703: I don't think it's a good idea. To implement it correctly, you have to fetch the key space configuration again in case someone else has also made some updates.

andylokandy: > you have to fetch the key space configuration again in case someone else has also made some updates

I can't see what conflicts with this.

proto/keyspace.proto
andylokandy: Because keyspace can have **many** prefixes and can bind/unbind prefix.

proto/pdpb.proto
BusyJay: What's point of returning `keyspace_version`?

PR #726, opened 2021-03-05
Revert "add receiver addr in the request header"

rleungx:
Reverts pingcap/kvproto#725


BusyJay (approved):
Reverting this because we use grpc metadata instead. And the field has never been used, so it can be removed safely.

nolouch (approved):

PR #728, opened 2021-03-07
coprocessor_v2: add new protocol messages

Jibbow:
Referring to new coprocessor issue https://github.com/tikv/tikv/issues/9747

Not sure about the naming here. Please let me know if I should use something like `coprv2` or anything else instead :)


skyzh (approved):

proto/coprocessor_v2.proto
```diff
@@ -0,0 +1,31 @@
...
+    string copr_version_constraint = 3;
+
+    bytes data = 4;
+}
+
+message RawCoprocessorResponse {
+    bytes data = 1;
+
+    errorpb.Error region_error = 2;
+    string other_error = 3;
```
sticnarf, line 30: It would be better to add a comment for this error. If the coprocessor is not found or the version is not compatible, the error will be returned in this field.

andylokandy: I think this kind of comment could be outdated very easily.

sticnarf: If there are some known error types, I think it would be better to specify where it is in the protocol.

andylokandy:
@Jibbow Seems it needs rebasing.

proto/coprocessor_v2.proto
Jibbow: Good point! Will add some known error messages as well.

Jibbow: Also, a bit unrelated: @andylokandy @skyzh should we use 
```
message Error {
  oneof error {
    ....
    string other_error = n;
  }
}

message RawCoprocessorResponse {
  oneof message {
    bytes data = 1;
    Error error = 2;
  }
}
```
to indicate that there can either be a successful result or an error?

andylokandy (approved):

andylokandy:
@ti-chi-bot wake up!

skyzh:
Seems that this repo is not controlled by not. We may need to ask those with write access to merge.

andylokandy:
@hi-rustin Call for father of the bot :lol:

andylokandy:
@skyzh Oh, really? I used to have the permission, and since ti-chi is launched the access is withdrawn...

hi-rustin:
This repository does not use tichi.

sticnarf (approved):

PR #730, opened 2021-03-08
Use GitHub Actions to run the release-3.0 tests

JmPotato:
Signed-off-by: JmPotato <ghzpotato@gmail.com>


JmPotato:
@nolouch @overvenus PTAL

disksing:
@JmPotato would you like to explain why do want it?

JmPotato:
> @JmPotato would you like to explain why do want it?

Well, this change is not necessary actually. I will close it.

PR #732, opened 2021-03-10
Add CAS method for RawKV

Little-Wallace:


proto/kvrpcpb.proto
```diff
@@ -1034,3 +1038,22 @@ message RawGetKeyTTLResponse {
...
+    // If this field does not exist, it means that the key-value should only put into database if and only if the key
+    // does not exist.
+    bytes previous_value = 4;
+    string cf = 5;
+    uint64 ttl = 6;
+}
+
+message RawCASResponse {
+    errorpb.Error region_error = 1;
+    string error = 2;
```
andylokandy, line 1055: What is this for?

proto/kvrpcpb.proto
```diff
@@ -1034,3 +1038,22 @@ message RawGetKeyTTLResponse {
...
+    bytes value = 3;
+    bool not_equals = 4;
```
andylokandy, line 1057: For consistency with L1039

```suggestion
    bool not_equal = 4;
```

proto/tikvpb.proto
```diff
@@ -50,6 +50,9 @@ service Tikv {
     // Get TTL of the key. 
     // If the key is non-TTL, returns 0
     rpc RawGetKeyTTL(kvrpcpb.RawGetKeyTTLRequest) returns (kvrpcpb.RawGetKeyTTLResponse) {}
+    // Put the key-value if and the previous value store in database equal RawCASRequest.previous_value. If not, return
+    // the previous value.
```
andylokandy, line 54:
```suggestion
    // Compare if the value in database equals to `RawCASRequest.previous_value` before putting the new value. If not, the mutation has no effect and the value in database will be returned.
```

andylokandy (approved):

proto/kvrpcpb.proto
```diff
@@ -446,6 +446,7 @@ message RawPutRequest {
     bytes value = 3;
     string cf = 4;
     uint64 ttl = 5;
+    bool for_cas = 6;
```
BusyJay, line 449: How about `check_cas`?

proto/kvrpcpb.proto
```diff
@@ -1034,3 +1038,22 @@ message RawGetKeyTTLResponse {
...
     bool not_found = 4;
 }
+
+message RawCASRequest {
+    Context context = 1;
+    bytes key = 2;
+    bytes value = 3;
+    // If this field does not exist, it means that the key-value should only put into database if and only if the key
+    // does not exist.
+    bytes previous_value = 4;
```
BusyJay, line 1048: Actually you can't tell non-exist value from empty value in protobuf.

Little-Wallace: If the value is a empty string, I think it present non-exist

proto/kvrpcpb.proto
Little-Wallace: I think there is no difference between them....

Connor1996 (approved):

proto/kvrpcpb.proto
andylokandy: Please don't do that. It'll be a legacy problem like this one:

https://github.com/tikv/tikv/issues/6711

andylokandy: I suggest adding a new bool field `should_not_exist` to indicate that property.

Connor1996 (commented):
You need to regenerate the files. There are a lot of changes after you change the version of protoc.

proto/kvrpcpb.proto
BusyJay: `for_cas` seems the key is put for CAS purpose. However, the field is meant to be participated in CAS checking.

proto/kvrpcpb.proto
```diff
@@ -1034,3 +1038,23 @@ message RawGetKeyTTLResponse {
     uint64 ttl = 3;
     bool not_found = 4;
 }
+
+message RawCASRequest {
+    Context context = 1;
+    bytes key = 2;
+    bytes value = 3;
+    // If the length of this field equals zero, it means that the key-value should only put into database if and only if the key
```
BusyJay, line 1046: Stale comment.

PR #735, opened 2021-03-12
add external storage error for backup

YuJuncen:
For a better fix for https://github.com/pingcap/br/issues/774, at backup site.

As the restore side, any errors in the stream response are for retryable download errors. (relatively, errors cannot be retried would send as an gRPC error response.)


proto/backup.proto
```diff
@@ -92,12 +92,23 @@ message ClusterIDError {
...
 }
 
+// Should we let this be an generic error?
+message ExternalStorageError {
+    bool retryable = 1;
+
+    // If use_backoff_time is true, 
+    // the client SHOULD back off for (backoff_time) millseconds.
+    bool use_backoff_time = 2;
+    uint32 backoff_time = 3;
```
kennytm, line 102: what the client should use if `retryable == true && use_backoff_time == false`?

it seems like the whole message should have one field: `uint32 retry_after`. 
* if zero, cannot retry
* if positive, can retry after ≥N milliseconds.

YuJuncen: Under this condition, the client can decide time time duration of backing off by itself. This design made `use_backoff_time` and `backoff_time` as a hint, yep, this might be redundant.

And IMO, `retry_after` as a int(negative for cannot retry), or with a `can_retry` would be better, for it can cover the case "retry immediately".

proto/backup.proto
```diff
@@ -92,12 +92,19 @@ message ClusterIDError {
     uint64 request = 2;
 }
 
+// Should we let this be an generic error?
+message ExternalStorageError {
+    bool retryable = 1;
+    uint32 retry_after = 2;
```
kennytm, line 98: add back the comment that the unit is millseconds.

kennytm (commented):
please undo the changes to the irrelevant *.go files.

rest LGTM

proto/backup.proto
YuJuncen, line 98:
```suggestion
    // when retry_after >= 0, the client SHOULD retry this request at least $retry_after milliseconds.
    uint32 retry_after = 2;
```

pkg/configpb/configpb.pb.go
```diff
@@ -62,7 +62,7 @@ func (x StatusCode) String() string {
 	return proto.EnumName(StatusCode_name, int32(x))
 }
 func (StatusCode) EnumDescriptor() ([]byte, []int) {
-	return fileDescriptor_configpb_0380a73eaf0b444c, []int{0}
```
kennytm, line 65: revert changes to these files

kennytm (changes requested):

pkg/configpb/configpb.pb.go
YuJuncen: Done, but how those files can be corrupted... 🤔

kennytm: well these random numbers are regenerated every time 🤔

kennytm (approved):
LGTM

YuJuncen:
friendly ping @overvenus

disksing (approved):

disksing:
Hi @YuJuncen Please resolve the conflict so I can merge it.

disksing:
close for inactivity, please reopen if need to merge.

PR #737, opened 2021-03-17
backup: add a CloudDynamic type

gregwebs:
This allows adding a cloud provider without
creating a new message type here.

These protobuf changes are tested in TiKV on my [cloud-storage branch](https://github.com/tikv/tikv/pull/9706). CloudDynamic has been tested for aws and gcp with the scli example.


proto/backup.proto
```diff
@@ -185,8 +187,27 @@ message GCS {
...
+    string region = 3;
+    string bucket = 4;
+    string prefix = 5;
+    string storage_class = 6;
+}
+
+// CloudDynamic allows adding new fields without changing protobuf definitions
+message CloudDynamic {
+    Bucket bucket = 1;
+    string name = 2;
```
kennytm, line 206:
```suggestion
    string service = 2;
```

`name` sounds like a file name here 🙃

proto/backup.proto
```diff
@@ -185,8 +187,27 @@ message GCS {
     // You can get one from https://console.cloud.google.com/apis/credentials, and
     // copy the content, set it as string here.
     string credentials_blob = 6;
+    Bucket bucket_info = 7;
 }
 
+// S3 and GCS can be transitioned to use this
+message Bucket {
+    string endpoint = 1; // use a different endpoint (host + url)
+    string host = 2; // use a different host
```
kennytm, line 196: why we need a `host` field

proto/backup.proto
```diff
@@ -185,8 +187,27 @@ message GCS {
...
+    string region = 3;
+    string bucket = 4;
+    string prefix = 5;
+    string storage_class = 6;
```
kennytm, line 200: i think `storage_class` belongs to the `attrs` field.

kennytm (commented):
I think it's better you deprecate `StorageBackend.{s3,gcs}` than pushing `bucket_info` into `S3` and `GCS`.

proto/backup.proto
gregwebs: how about `provider_name` or `cloud_provider` or `cloud_name` ?

proto/backup.proto
gregwebs: For GCS it is easier to give a host `localhost:4443` rather than having to know the full url of `localhost:4443/storage/v1`

proto/backup.proto
gregwebs: If anything is expected to be common between clouds I put it in Bucket. Both AWS and GCP support specifying a storage class so I expect everyone else to as well. Maybe it is not available in many clouds?

gregwebs:
> I think it's better you deprecate StorageBackend.{s3,gcs} than pushing bucket_info into S3 and GCS.

I can see the advantage that there are no duplicated fields. However, I already implemented checking for both fields in TiKV so this is then only really an advantage for the caller.

Switching to two separate types means 2 separate APIs will now be available? I can see the appeal to switching the url and not having deprecated fields. However I think the point of protobufs is that this can be avoided as long as the message body is evolving in a backwards-compatible way. The client will have to determine the proper TiKV version and choose between 2 urls. Whereas now the client can fill out all the fields (in duplicate) without worrying about the TiKV version.
And again, it is more work in TiKV as well since it is already implemented to work as is.

proto/backup.proto
kennytm: also good

proto/backup.proto
kennytm: this can be manipulated in the user interface, no need to be represented the gRPC protocol.

kennytm:
> I can see the advantage that there are no duplicated fields. However, I already implemented checking for both fields in TiKV so this is then only really an advantage for the caller.

I could say keeping the current situation is "only really an advantage for the callee" 😉. 

And actually it is possible to just convert the S3 / GCS structure into CloudDynamic inside TiKV before forwarding to the cloud crate. 

> The client will have to determine the proper TiKV version and choose between 2 urls. 

well in the current implementation the client always have to choose S3 / GCS over CloudDynamic too since it is a `oneof` in StorageBackend 😞. We could change the message instead to

```protobuf
message StorageBackend {
    oneof backend {
        Noop noop = 1;
        Local local = 2;
        S3 s3 = 3 [deprecated = true];
        GCS gcs = 4 [deprecated = true];
    }
    CloudDynamic cloud_dynamic = 5;
}
```

but then for this we could keep the messages S3 and GCS untouched.

proto/backup.proto
gregwebs: I don't know what you mean. Do you mean to give just a host as the endpoint and have TiKV figure out the intent?

gregwebs:
> And actually it is possible to just convert the S3 / GCS structure into CloudDynamic inside TiKV before forwarding to the cloud crate.

Yes, I have conversions to/from CloudDynamic on my branch.

> We could change the message instead to

Do you mean to add a field `Bucket` to StorageBackend? And perhaps the attrs? Moving CloudDynamic out of the oneof would require more code to merge fields and I think look confusing as to what should be sent in the message. The main idea behind CloudDynamic is that we don't need to alter protobuf definitions when adding support for entirely new cloud vendors. While making these changes I realized there should be a common Bucket type.

proto/backup.proto
kennytm: let's say BR provides both `--gcs.endpoint` and `--gcs.host` parameters. then BR will send

```go
endpoint := gcs.endpoint
if endpoint == "" {
    endpoint = gcs.host + "/storage/v1"
}
```

to TiKV.

But for GCS... I thought the endpoint URL is *still* `http://localhost:4443`? The GCS library should append the `/storage/v1` part itself when constructing the URL.

gregwebs: The endpoint url needs /storage/v1, at least for fake-gcs-server. TiKV does not automatically add it. I specifically tried this and then wrote unit tests to verify the current behavior. I didn't want to make a breaking change in behavior so I created a new field.

I agree that the code you added to BR makes sense given what I know. My concerns are:
* Is there some use case where storage/v1 should not be added to the url? What happens when TiKV switches to storage/v2 ?
* external_storage can be used for more than just backup in the future. TiKV or PD could be storing potentially any of their data to external_storage. Physical backup could be run automatically from TiKV without an external tool.

tennix:
@lcwangchao PTAL

gregwebs:
oh, now I see the deprecation marker. I don't know if we actually want to deprecate S3/GCS. It is nice to have fields explicitly listed (things are self-documenting) rather than be a hashmap. CloudDynamic is designed to support future new cloud providers without requiring alteration of protobuf definitions.

proto/backup.proto
gregwebs: okay, I changed it to "provider_name"

proto/backup.proto
kennytm: Ok. Well the situation is not good...

* Currently GCS + custom endpoint is not well tested on BR (pingcap/br#258), so I hope we are allowed to break compatibility for this experimentally if a more meaningful choice can be taken 🙃.

* > Is there some use case where storage/v1 should not be added to the url? 

  Maybe a custom GCS service won't bother to attach the storage/v1 part, dunno. If there is a storage/v2 then it makes more sense to provide the "host" version of endpoint only.

* > TiKV or PD could be storing potentially any of their data to external_storage. 

  TiKV and PD and do the host → endpoint URL normalization too.

proto/backup.proto
kennytm: I just think that endpoint/host, region, bucket, prefix are all essential for locating the resource, so having storage_class here a bit strange 😅.

proto/backup.proto
```diff
@@ -143,6 +143,7 @@ message StorageBackend {
         Local local = 2;
         S3 s3 = 3;
         GCS gcs = 4;
+	CloudDynamic cloud_dynamic = 5;
```
kennytm, line 146:
```suggestion
        CloudDynamic cloud_dynamic = 5;
```

kennytm (commented):
There's still the question of host vs endpoint.

LGTM other than these.

proto/backup.proto
gregwebs: I tested the GCS endpoint flag against gcs-fake-server and was able to get it to work. The command was written down here: https://github.com/tikv/tikv/pull/9706
I also tested setting the host with the scli example.

If we look at the [GCS storage API](https://cloud.google.com/storage/docs/request-endpoints), there are multiple url endpoints.

* https://storage.googleapis.com/storage/v1/PATH_TO_RESOURCE
* https://storage.googleapis.com/upload/storage/v1/b/BUCKET_NAME/o
* https://storage.googleapis.com/batch/storage/v1/PATH_TO_RESOURCE
* https://storage.googleapis.com/download/storage/v1/b/BUCKET_NAME/o/OBJECT_NAME?alt=media

The TiKV code has 2 of them now. To me it doesn't make sense to be giving one of the urls as the endpoint. Actually I think this code will already fail if a user gives storage/v1 but the API needs to use a different url.
What makes sense to me is to that the endpoint is the common subset of urls. Effectively the host flag does this now. I didn't want to alter the existing behavior of the endpoint field and break backwards compatibility.

proto/backup.proto
gregwebs: Maybe I would call your concept a BucketLocation. I would much prefer to not use attrs for commonly used fields. attrs can have typos and are more difficult to document.

proto/backup.proto
gregwebs: [This test case in my PR](https://github.com/tikv/tikv/pull/9706/files#diff-60ec9b1be5e8a5a1732ccbab6e3c6aa7d1643f059bd1347edca5e62d236eb687R510) demonstrates that the system of overriding the full endpoint url is broken. That is it always changes the full url even though there are 2 different urls that should be in use.

That PR now has more merge conflicts to resolve because I am waiting on this PR to be merged. Can we come to a resolution on this conversation?

kennytm: I'll send this to the team for discussion tomorrow. My preference:

* `--gcs.endpoint` doesn't need to care about backward compatibility since it's not documented at all
* the endpoint should have the same behavior as https://pkg.go.dev/google.golang.org/api/option#WithEndpoint. AFAIK both of the popular Rust GCS crates `tame_gcs` and `cloud-storage` hardcoded the endpoint URL 😢
* If `--gcs.host` and `--gcs.endpoint` are different by default, the former should be resolved to the endpoint URL at the client library level. I prefer not to have both the host and endpoint fields in BucketInfo, since it is unnecessary for S3. Client-side normalization is also how we handle `--s3.provider` vs "force-path-style".

gregwebs: For `S3` I don't have the host field, only for `Gcs` and CloudDynamic`. We could remove host from `Bucket` used by `CloudDynamic` as well. 

I would be happy to change the behavior of `gcs.endpoint`

Note that it is possible (although obviously unwieldy) when using the client libraries to create a client per url. I suspect that they are both poorly designed and almost nobody tries to use that option.

kennytm: We've discussed about the compatibility issue for BR, and the conclusion is that:

"BR GCS support is still not released as stable. Therefore, we accept incompatible changes."

----

> Note that it is possible (although obviously unwieldy) when using the client libraries to create a client per url

yeah we support both the parameter style `-s 'gcs://foo/bar' --gcs.endpoint https://1.2.3.4:5678` and URL style `-s 'gcs://foo/bar?endpoint=https://1.2.3.4:5678'`. The same for `host` if it exists.

gregwebs: So we should change the behavior of `endpoint` for GCS to not have /storage/v1? Then we don't need the `host` field.

kennytm: > So we should change the behavior of `endpoint` for GCS to not have /storage/v1

SGTM.

gregwebs: okay, that change is on this PR now and I will make the change on the TiKV PR as well.

proto/backup.proto
kennytm: @gregwebs this went back to tabs again 😅

lcwangchao (approved):

kennytm (approved):

proto/backup.proto
```diff
@@ -187,6 +188,23 @@ message GCS {
...
+    string bucket = 4;
+    string prefix = 5;
+    string storage_class = 6;
+}
+
+// CloudDynamic allows adding new fields without changing protobuf definitions
+message CloudDynamic {
+    Bucket bucket = 1;
+    string provider_name = 2;
+    map<string, string> attrs = 3;
```
yiwu-arbug, line 204: what is being held in the map? Can we define it as a struct (protobuf message) so that the fields are well defined?

proto/backup.proto
```diff
@@ -187,6 +188,23 @@ message GCS {
...
+    string region = 3;
+    string bucket = 4;
+    string prefix = 5;
+    string storage_class = 6;
+}
+
+// CloudDynamic allows adding new fields without changing protobuf definitions
+message CloudDynamic {
+    Bucket bucket = 1;
+    string provider_name = 2;
```
yiwu-arbug, line 203: I know we cannot define constant in protobuf. But can this be an enum? or at least inline comment what are the possible provider names.

gregwebs: I can comment that s3/gcs are available. However the entire point of this is that it is dynamic and that we don't have to alter all of our code paths to experiment with new cloud providers.

yiwu-arbug: I'm fine just comment what are the possible values. But I think the purpose of protobuf is to make things well-defined. Otherwise we could use JSON or XML just for flexibility.

gregwebs: Protobuf supports dynamic maps and also arrays of dynamic length via repeat. XML has always supported strong schemas and these are available for JSON as well now. I think the main reason for Protobuf is a more efficient encode/decode.
Most of the time things should be well-defined. Occasionally we want things to be more dynamic. Just as usually we are happy with a strong schema in a database but now we support JSON as well for the times where it is useful.

proto/backup.proto
gregwebs: I added a comment about the purpose of CloudDynamic. The fields in Bucket are common, but otherwise every cloud provider can have arbitrary attributes.

overvenus (approved):
LGTM

PR #738, opened 2021-03-18
*:add scan_used_time_ms for TimeDetail

jyz0309:
Signed-off-by: jyz0309 <45495947@qq.com>

add scan_used_time_ms for TimeDetail 

scan_used_time_ms: The time used in TiKV scan rows


jyz0309:
@crazycs520 @breeswish PTAL

proto/kvrpcpb.proto
```diff
@@ -863,6 +863,8 @@ message TimeDetail {
     // cannot be excluded for now, like Mutex wait time, which is included in this field, so that
     // this field is called wall time instead of CPU time.
     int64 process_wall_time_ms = 2;
+    // Scan Used Time means the time used in TiKV scan.
```
crazycs520, line 866:
```suggestion
    // Scan Used Time means the time used in key/value scan.
```

crazycs520: TiFlash also has scan duration.

proto/kvrpcpb.proto
```diff
@@ -863,6 +863,8 @@ message TimeDetail {
...
+    int64 scan_used_time_ms =3;
```
breeswish, line 867:
```suggestion
    int64 kv_read_wall_time_ms = 3;
```

breeswish: I think calling it read is better, since read operation includes get and scan, and actually we want time measurements for both.

breeswish (approved):

crazycs520 (approved):

proto/kvrpcpb.proto
```diff
@@ -863,6 +863,8 @@ message TimeDetail {
     // cannot be excluded for now, like Mutex wait time, which is included in this field, so that
     // this field is called wall time instead of CPU time.
     int64 process_wall_time_ms = 2;
+    // KV read wall Time means the time used in key/value scan and get.
+    int64 kv_read_wall_time_ms = 3;
```
zhongzc, line 867: What's the difference between `process_wall_time_ms` and `kv_read_wall_time_ms`?

Is `kv_read_wall_time_ms` with same semantics as `process_wall_time_ms`, but for special kind of read requests, i.e. `get` and `batch_get`?

breeswish: In coprocessor they are different.

PR #740, opened 2021-03-22
Revert and delete kv_read_wall_time_ms

jyz0309:
Revert and delete kv_read_wall_time_ms


jyz0309:
@crazycs520 @breeswish PTAL

breeswish:
What if it is a read from KvGet or KvBatchGet?

PR #741, opened 2021-03-23
Use a docker image to generate protobufs

gregwebs:
Protobufs are tricky to install, especially for multiple languages.
But we already have a reliable install in our github workflows.

Build a docker image based of the github workflows.

This is tested to work for Go and Rust. For C++ it makes it through without error but it doesn't seem to be updating the generated code. Perhaps someone familiar with the C++ build can contribute on top of this PR.


Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM ubuntu:20.04
```
BusyJay, line 1: How about using namely/docker-protoc to save a lot of build time?

BusyJay: And setting environment GRPCIO_SYS_USE_PKG_CONFIG to 1 will force rust to link to share library grpc too, which can also save build time further.

gregwebs: I tried namely/docker-protoc previously and ran into complications. This PR matches our github workflows which made it relatively easy to get working.

gregwebs: I tried adding the env variable you recommend and then the build errors with ` thread 'main' panicked at 'can't find library grpc via pkg-config`. I am quite happy with this PR as is but would welcome future PRs that improve things further.

gregwebs: A way to save docker image build time the best way would be to publish this image on dockerhub. I don't know who to contact about that.

BusyJay (approved):

overvenus (approved):
LGTM

PR #742, opened 2021-03-25
Update RawCAS

andylokandy:

Refined the name of CAS from `compare and set` to `compare and swap` according to the [definition of CAS](https://en.wikipedia.org/wiki/Compare-and-swap) where `compare and swap` returns the current value when comparison is failed and `compare and set` only returns a bool indicating the failure.

`RawCASRespond` is also changed to reflect the sematic.


ekexium:
LGTM

ekexium (approved):

Little-Wallace (approved):
LGTM

PR #743, opened 2021-03-25
storage: Add checksum/split-region/import-write for rawkv

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>

We want to support loading data into TiKV in RawKV mode by tidb-lightning. So we need to supply these method for tidb-lightning:

- RawWrite: to write data with RawKV-format rather than TxnKV-format.
- RawSplitRegion: tidb-lightning will split region into multiple ones to avoid write too many data in one region.
- RawChecksum: tidb-lightning will compare checksum-value between source data and data in TiKV. But TiKV used to supply TxnKV method for checksum and the checksum code is in `tidb` repo. So we need an independent method only for RawKV.


nolouch:
RawSplitRegion-> what's the differnet with https://github.com/pingcap/kvproto/pull/743/files#diff-9012f74300ab5335dd6edf75bb4f95361c48f5454d4a0fb4c51864989ee2f3bfR89-R91？

Little-Wallace:
> RawSplitRegion-> what's the differnet with [https://github.com/pingcap/kvproto/pull/743/files#diff-9012f74300ab5335dd6edf75bb4f95361c48f5454d4a0fb4c51864989ee2f3bfR89-R91？](https://github.com/pingcap/kvproto/pull/743/files#diff-9012f74300ab5335dd6edf75bb4f95361c48f5454d4a0fb4c51864989ee2f3bfR89-R91%EF%BC%9F)

Because the keys of request to rawsplitregion will not be encoded.

disksing:
Hi @Little-Wallace @nolouch  do we need to continue it?

disksing:
closing for inactivity

PR #745, opened 2021-03-30
sst: add total bytes and total kvs into SSTMeta

3pointer:
Part of https://github.com/pingcap/br/issues/964

We need to add two fields `total_bytes` and `total_kvs`. so that TiKV can set size and keys correctly to PD through heartbeat.


glorv:
LGTM

glorv (approved):

Little-Wallace (approved):
LGTM

PR #747, opened 2021-03-31
Add read byte in ScanDetailV2 for future

jyz0309:
Add read byte in ScanDetailV2 for future


proto/kvrpcpb.proto
```diff
@@ -894,22 +893,25 @@ message ScanDetailV2 {
     // there are a lot of deleted MVCC keys.
     uint64 total_versions = 2;
 
+    // total number of bytes from Scan or kvGet
+    uint64 read_bytes = 3;
+
     // Total number of deletes and single deletes skipped over during
     // iteration, i.e. how many RocksDB tombstones are skipped.
-    uint64 rocksdb_delete_skipped_count = 3;
+    uint64 rocksdb_delete_skipped_count = 4;
```
breeswish, line 901: This is a breaking change and please don't do it. 

https://developers.google.com/protocol-buffers/docs/proto#assigning_field_numbers

> each field in the message definition has a unique number. These numbers are used to identify your fields in the message binary format, and should not be changed once your message type is in use.

proto/kvrpcpb.proto
```diff
@@ -894,6 +893,9 @@ message ScanDetailV2 {
     // there are a lot of deleted MVCC keys.
     uint64 total_versions = 2;
 
+    // Total number of bytes from Scan or kvGet
+    uint64 read_bytes = 8;
```
crazycs520, line 897: since the index number is `8`, please put after line 914.

breeswish (approved):

crazycs520:
Wait for the related TiDB/TiKV PR to merge them together at the same time.

breeswish (approved):

breeswish (approved):

proto/kvrpcpb.proto
```diff
@@ -815,6 +814,9 @@ message ScanDetailV2 {
 
     // Total number of bytes from block reads.
     uint64 rocksdb_block_read_byte = 7;
+
+    // Total number of bytes from Scan or kvGet
```
breeswish, line 818: The comment does not match the field name, how about changing it to "Total number of bytes read from RocksDB"?

BTW what is the behaviour when it comes to tombstones or deleted (but not gced) multi version keys?

jyz0309: Done

proto/kvrpcpb.proto
```diff
@@ -815,6 +814,9 @@ message ScanDetailV2 {
 
     // Total number of bytes from block reads.
     uint64 rocksdb_block_read_byte = 7;
+
+    // Total number of bytes read from RocksDB
```
breeswish, line 818: Considering our latest spec, maybe we should change the comment?

jyz0309: Of course, I will change it before dev :)

PR #748, opened 2021-04-06
add definitions for an ExternalStorage service

gregwebs:
This PR is required for this TiKV PR: https://github.com/tikv/tikv/pull/9706

The protobuf definitions here are usable for either a GRPC service or for calling to dynamically linked code built as a C library where the arguments to the C function are bytes from a protobuf (in this case the service definition isn't used). 

Its actual usage in that PR for either purpose requires turning on a feature flag.


proto/backup.proto
```diff
@@ -214,3 +214,27 @@ message BackupResponse {
...
+    uint64 content_length = 3;
+}
+
+message ExternalStorageWriteResponse {
+}
+
+service ExternalStorage {
+    rpc restore(ExternalStorageRestoreRequest) returns (ExternalStorageRestoreResponse) {}
+    rpc write(ExternalStorageWriteRequest) returns (ExternalStorageWriteResponse) {}
+}
```
overvenus, line 240: Could you add some comments about the service and PRCs?

gregwebs: comments added

proto/backup.proto
```diff
@@ -214,3 +214,32 @@ message BackupResponse {
 service Backup {
     rpc backup(BackupRequest) returns (stream BackupResponse) {}
 }
+
+// ExternalStorage is a service for using a cloud backend from StorageBackend to store files.
+// This can be used to backup and restore SST files.
+service ExternalStorage {
+    // Restore to a file
+    rpc restore(ExternalStorageRestoreRequest) returns (ExternalStorageRestoreResponse) {}
```
overvenus, line 222: Why not name it `read`?

gregwebs: ExternalStorage in TiKV has a "read" method that creates a read stream from S3. The restore method goes a step further and saves it to a file.

The direction that would make more sense to me would be to instead call this write API (which takes a file, not a stream) something like "save". More descriptive would be "save file to S3"

overvenus (approved):
LGTM

proto/backup.proto
```diff
@@ -214,3 +214,32 @@ message BackupResponse {
...
+    StorageBackend storage_backend = 1;
+    string object_name = 2;
+    string restore_name = 3;
+    uint64 content_length = 4;
+}
+
+message ExternalStorageRestoreResponse {
+}
+
+message ExternalStorageWriteRequest {
```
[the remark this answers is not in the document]
gregwebs: thanks, I changed that now.

kennytm (approved):

PR #751, opened 2021-04-12
mpp: support returning regions that need retry

hanfei1991:
Mpp need to refresh the TiDB client's cache when there are region retry in TiFlash. TiFlash returns these region info and TiDB refresh the cache.


windtalker (approved):
LGTM

fzhedu (approved):
LGTM

PR #752, opened 2021-04-22
Add proto for implementing lock view

MyonKeminta:
The SQL information will be sent via the `Context.resource_group_tag` field, which may also be made use of by another feature *Top SQL*, cc @breeswish


breeswish (approved):

longfangsong (approved):
/lgtm

youjiali1995 (approved):

PR #757, opened 2021-04-29
add store status into store heartbeat response

hicqu:


proto/pdpb.proto
```diff
@@ -516,10 +516,17 @@ message StoreHeartbeatRequest {
     StoreStats stats = 2;
 }
 
+enum StoreStatus {
```
nolouch, line 519: How about adding it in https://github.com/pingcap/kvproto/blob/7d50a22e4b4fb0da7e003ac72d639711cf3ead5e/proto/metapb.proto#L23-L27

PR #759, opened 2021-04-30
Add missing fields to mvcc info

sticnarf:
After async commit is introduced, there are a lot more fields adding to the mvcc info. For easy diagnosis, we should add these fields to the proto.


youjiali1995 (approved):

disksing:
Do we need to proceed with this pull request? cc @sticnarf @youjiali1995

disksing (approved):

PR #762, opened 2021-05-11
Support DuplicateDetect for lightning-import

Little-Wallace:
Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>


proto/import_sstpb.proto
```diff
@@ -221,3 +224,19 @@ message WriteResponse {
     Error error = 1;
     repeated SSTMeta metas = 2;
 }
+
+message DuplicateDetectRequest {
+    kvrpcpb.Context context = 1;
+    bytes start_key = 2;
+    // Return only the keys found by scanning, not their values.
```
gozssky, line 231: This line should be moved before `key_only`?

proto/import_sstpb.proto
```diff
@@ -221,3 +224,19 @@ message WriteResponse {
...
+    bytes end_key = 3;
+    bool key_only = 4;
+    // We only check the data whose timestamp is larger than `min_commit_ts`.
+    uint64 min_commit_ts = 5;
```
gozssky, line 235: Is it better that also check data including `min_commit_ts` not only `large than`

Little-Wallace: Emmmm. Maybe not include.....

proto/import_sstpb.proto
```diff
@@ -51,6 +51,9 @@ service ImportSST {
     
     // Ingest Multiple files in one request
     rpc MultiIngest(MultiIngestRequest) returns (IngestResponse) {}
+
+    // Upload an SST file to a server.
```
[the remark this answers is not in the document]
Little-Wallace: Yes... I will remove it

proto/import_sstpb.proto
```diff
@@ -221,3 +224,19 @@ message WriteResponse {
     Error error = 1;
     repeated SSTMeta metas = 2;
 }
+
+message DuplicateDetectRequest {
+    kvrpcpb.Context context = 1;
+    bytes start_key = 2;
+    // Return only the keys found by scanning, not their values.
+    bytes end_key = 3;
+    bool key_only = 4;
```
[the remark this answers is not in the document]
Little-Wallace: For a handle key, we do not need to calculate the value at once, because we may get a repeated handle when detecting other unique key.

proto/import_sstpb.proto
```diff
@@ -221,3 +224,25 @@ message WriteResponse {
...
+message KvPair {
+    bytes key = 1;
+    bytes value = 2;
+    uint64 commit_ts = 3;
+}
+
+message DuplicateDetectResponse {
+    errorpb.Error region_error = 1;
+    Error key_error = 2;
+    repeated KvPair pairs = 3;
```
kennytm, line 247: please add a comment about how the kv pairs are arranged e.g.

```
[
  {key: "key1", value: "value11", commit_ts: 1001},
  {key: "key1", value: "value12", commit_ts: 1004},
  {key: "key1", value: "value13", commit_ts: 1005},
  {key: "key2", value: "value21", commit_ts: 1002},
  {key: "key2", value: "value22", commit_ts: 1004},
  ...
]
```

kennytm (approved):
lgtm, but a minor nit

proto/import_sstpb.proto
kennytm: explicitly mention that `min_commit_ts` is "excluded"?

proto/import_sstpb.proto
Little-Wallace: OK.

gozssky: Is the  pairs' order guaranteed？How about add a comment to explain it?

Little-Wallace: Yes. It will be returned in order, but I don’t think we should rely on this assumption

gozssky (approved):

PR #763, opened 2021-05-13
mpp: support returning regions that need retry (#751)

hanfei1991:
* mpp: support return of region errors

* return region list


fzhedu (approved):
LGTM

windtalker (approved):
LGTM

PR #773, opened 2021-05-31
Add ResourceMetering agent service definition

dragonly:
Add Top SQL agent service and message definitions.
This is used to send Top SQL records to an agent for persistent storage.

Note that TiKV should not be aware of SQL related concepts, so this is actually a general resource metrics collecting agent, which can be used by TiDB to pass SQL related tags, encoded in binary payload.

related to https://github.com/pingcap/tipb/pull/225


proto/topsql_agent.proto
```diff
@@ -0,0 +1,26 @@
...
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (rustproto.lite_runtime_all) = true;
+
+option java_package = "org.tikv.kvproto";
+
+// TopSQLAgent is the persistent agent service for TopSQL records
+service TopSQLAgent {
```
breeswish, line 17: TiKV should not know anything about "SQL".

proto/topsql_agent.proto
```diff
@@ -0,0 +1,26 @@
...
+  // CollectTikv is called periodically (e.g. per minute) to save the in-memory TopSQL records
+  rpc CollectTikv(stream CPUTimeRequestTikv) returns (google.protobuf.Empty) {}
```
breeswish, line 19:
```suggestion
  rpc CollectTikvMeteringResult(stream ResourceMeteringResult) returns (CollectTikvMeteringResultResponse) {}
```

proto/topsql_agent.proto
```diff
@@ -0,0 +1,26 @@
...
+}
+
+message CPUTimeRequestTikv {
+  repeated uint64 timestamp_list = 1;  // timestamp in second
+  repeated uint32 cpu_time_ms_list = 2;  // this can be greater than 1000 when counting concurrent running SQL queries
+  bytes payload = 3;  // the specific encoding is opaque to tikv-server
```
breeswish, line 25: What is payload? Is it ResourceTag?

dragonly: Yes, I should name it more specifically.

Cargo.toml
```diff
@@ -12,12 +12,13 @@ path = "src/lib.rs"
...
 default = ["protobuf-codec"]
 protobuf-codec = ["protobuf-build/grpcio-protobuf-codec", "raft-proto/protobuf-codec", "grpcio/protobuf-codec"]
-prost-codec = ["prost", "prost-derive", "lazy_static", "protobuf-build/grpcio-prost-codec", "raft-proto/prost-codec", "grpcio/prost-codec"]
+prost-codec = ["prost", "prost-derive", "prost-types", "lazy_static", "protobuf-build/grpcio-prost-codec", "raft-proto/prost-codec", "grpcio/prost-codec"]
 
 [dependencies]
 protobuf = "=2.8.0"
 prost = { version = "0.7", optional = true }
 prost-derive = { version = "0.7", optional = true }
+prost-types = { version = "0.7", optional = true }
```
breeswish, line 21: This can be reverted

proto/topsql_agent.proto
```diff
@@ -0,0 +1,27 @@
...
+option (gogoproto.sizer_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (rustproto.lite_runtime_all) = true;
+
+option java_package = "org.tikv.kvproto";
+
+// ResourceUsageAgent is the persistent agent service for resource usage records
+service ResourceUsageAgent {
+  // CollectCPUTime is called periodically (e.g. per minute) to save the in-memory resource usage records
+  rpc CollectCPUTime(stream CPUTimeRequestTikv) returns (CPUTimeResponseTikv) {}
```
breeswish, line 18:
```suggestion
  rpc CollectCPUTime(stream CollectCPUTimeRequest) returns (CollectCPUTimeResponse) {}
```

proto/topsql_agent.proto
```diff
@@ -0,0 +1,27 @@
...
+}
+
+message CPUTimeRequestTikv {
+  repeated uint64 timestamp_list = 1;  // timestamp in second
+  repeated uint32 cpu_time_ms_list = 2;  // this can be greater than 1000 when counting concurrent running requests
+  bytes resource_tag = 3;  // the specific encoding is opaque to tikv-server
```
breeswish, line 24:
```suggestion
  bytes resource_tag = 3;  // the tag of the resource, by which CPU time is grouped.
```

BTW since `resource_tag` is such important (as a Primary Key) I suggest to put it as the first field.

breeswish (approved):

zhongzc (approved):

PR #775, opened 2021-06-07
Better integration with cmake

JaySon-Huang:
Automatically re-generate `.pb.h` and `.pb.cc` if `.proto` file changed, don't need to run `scripts/generate_cpp.sh` manually.


windtalker:
LGTM

leiysky (approved):
LGTM

.gitignore
```diff
@@ -16,6 +16,7 @@ _tools
 cpp/
 !cpp/CMakeLists.txt
 proto-cpp/
```
tisonkun, line 18:
```suggestion
```

JaySon-Huang: We still provide the script `scripts/generate-cpp.sh` for those project don't use CMake. I think we can keep this.

tisonkun: Well.

tisonkun (approved):
LGTM. Though it would be better if we wrapper actions into a function, as the official `FindProtobuf.cmake` does.

PR #778, opened 2021-06-21
copr: add `processed_versions_size` to ScanDetailV2

zhongzc:
Signed-off-by: Zhenchi <zhongzc_arch@outlook.com>


proto/kvrpcpb.proto
```diff
@@ -815,6 +815,12 @@ message ScanDetailV2 {
 
     // Total number of bytes from block reads.
     uint64 rocksdb_block_read_byte = 7;
+
+    // Total number of keys from the scan executor (i.e. TableScan or IndexScan) reads.
+    uint64 scan_executor_read_keys = 8;
```
breeswish, line 820: how about non executors (like KvScan)?

proto/kvrpcpb.proto
```diff
@@ -815,6 +815,12 @@ message ScanDetailV2 {
 
     // Total number of bytes from block reads.
     uint64 rocksdb_block_read_byte = 7;
+
+    // Total number of keys from the scan executor (i.e. TableScan or IndexScan) reads.
```
breeswish, line 819: BTW this value may be exactly the same as `processed_versions` (please confirm).

zhongzc: ur right

zhongzc:
Updated. cc @breeswish

breeswish (approved):

jyz0309 (approved):

proto/kvrpcpb.proto
zhongzc: Found that `ScanResponse` doesn't have statistic fields,

https://github.com/pingcap/kvproto/blob/706fcaf286c8dd07ef59349c089f53289a32ce4c/proto/kvrpcpb.proto#L64-L72

is there a plan to add some?

PR #781, opened 2021-07-06
mpp: add proto to check if tiflash node is alive

hanfei1991:


proto/mpp.proto
```diff
@@ -19,6 +19,12 @@ message TaskMeta {
 	string address = 4; // target address of this task.
 }
 
+message IsAliveRequest {
+}
+
+message IsAliveResponse {
+}
```
windtalker, line 26: Should TiFlash send the response if it is not down, but it can not handle mpp request(for example, the server is in terminating state)?

hanfei1991: It's a little complicated. I suggest leave it like this.

windtalker (approved):
LGTM

fzhedu (approved):
LGTM

tisonkun:
@hanfei1991 is there any further information why we change the protocol?

PR #788, opened 2021-07-27
*: solve cloud tikv disk full problem

tier-cap:
### What problem does this PR solve?

Issue Number：https://github.com/tikv/tikv/issues/10537

When the disk is full, no matter one or two or the majority or all, what we expected will be:

1. TiKV not crash directly.
2. Support restoring storage through drop table/delete from table, expansion is also allowed.
3. Provide slightly better reading ability.

###  What is changed and how it works?

1. Add operations options when TiKV disk full happens.
2. Add disk full error code.

### Release note

- No release note.


proto/kvrpcpb.proto
```diff
@@ -686,9 +686,9 @@ message Context {
     // Any additional serialized information about the request.
     bytes resource_group_tag = 18;
 
-    // Flag used to tell tikv that operations can be allowed even when tikv disk full happens.
-    bool allowed_on_disk_full = 19;
-
+    // Used to tell TiKV what operations are allowed when the disk status is at different levels.
+    AllowedLevel allowed_level = 19;
```
disksing, line 690: This is too ambiguous. We can tell what is it used for from the name.

disksing: Maybe consider `DiskFullOpt`

proto/kvrpcpb.proto
```diff
@@ -766,6 +766,13 @@ enum IsolationLevel {
     RC = 1; // RC = read committed
 }
 
+//Operation allowed level during each TiKV disk full threshold.
+enum AllowedLevel {
+    Allowed_Normal = 0; // Normal is the default value, meaning all operations are allowed when disk status is normal.
```
hicqu, line 771: It's better to be `Normal`, `AllowedOnAlmostFull` and `AllowedOnAlreadyFull`.

proto/kvrpcpb.proto
```diff
@@ -766,6 +766,13 @@ enum IsolationLevel {
     RC = 1; // RC = read committed
 }
 
+//Operation allowed level during each TiKV disk full threshold.
```
hicqu, line 769: Generally there should be a space after `//`.

disksing:
@tier-cap  want to update the naming?

proto/kvrpcpb.proto
tier-cap: changed

proto/kvrpcpb.proto
tier-cap: changed

proto/kvrpcpb.proto
```diff
@@ -766,6 +766,13 @@ enum IsolationLevel {
     RC = 1; // RC = read committed
 }
 
+// Operation allowed info during each TiKV storage threshold.
+enum DiskFullOpt {
+    AllowedOnNormal = 0; // the default value, means all operations are allowed when disk status is normal.
```
5kbpers, line 771:
```suggestion
    Normal = 0; // the default value, means all operations are allowed when disk status is normal.
```

5kbpers (commented):
rest LGTM

proto/kvrpcpb.proto
tier-cap: CommandPri::Normal

hicqu: I prefer `Normal` too. I think the compile error is not related to this change.

hicqu: I know why `Normal` is bad. So please change to `Default`.

5kbpers (approved):

proto/kvrpcpb.proto
tier-cap: ok, done!

disksing (approved):

disksing:
@tier-cap Please resolver conflicts so I can merge it.

proto/kvrpcpb.proto
tier-cap: done!

hicqu (approved):

PR #791, opened 2021-08-05
report disk usage in raft message

hicqu:


proto/errorpb.proto
```diff
@@ -23,7 +23,7 @@ message NotLeader {
 
 message DiskFull {
     // The requested store ID
-    uint64 store_id = 1;
+    repeated uint64 store_id = 1;
```
disksing, line 26: why should we use repeated?

pkg/kvrpcpb/kvrpcpb.pb.go
```diff
@@ -53,7 +53,7 @@ func (x CommandPri) String() string {
 	return proto.EnumName(CommandPri_name, int32(x))
 }
 func (CommandPri) EnumDescriptor() ([]byte, []int) {
-	return fileDescriptor_kvrpcpb_da15c4d6242f2aec, []int{0}
+	return fileDescriptor_kvrpcpb_e53caf779f8a2335, []int{0}
```
disksing, line 56: Looks not related changes. Better to reset these files.

proto/errorpb.proto
hicqu: The Region leader can be ok but majority peers are in disk full states. This case can be detected by the leader and a DiskFull error will be returned.

pkg/kvrpcpb/kvrpcpb.pb.go
hicqu: In fact I change a comment in `proto/kvrpcpb.proto`. I think it's caused by that. So how about keeping it because other guys can also meet this change after he runs `make go`.

disksing: ok. It seems that the sha code are random generated...

disksing (approved):

5kbpers (approved):

PR #792, opened 2021-08-16
Add `has_locks` field in ScanResponse

ekexium:
A hotfix for TiSpark.


proto/kvrpcpb.proto
```diff
@@ -131,6 +131,7 @@ message KvPair {
 message ScanResponse {
     errorpb.Error region_error = 1;
     repeated KvPair pairs = 2;
+    bool has_locks = 3;
```
cfzjywxk, line 134: We may need to keep it compatible with the current [version](https://github.com/pingcap/kvproto/blob/master/proto/kvrpcpb.proto#L71),after the hotfix patch, they will upgrade to v5.0 in the future. The tikv server will be replaced and restarted first and it's possible the v3.0 hotfix tispark client reads the v5.0 tikv server.

cfzjywxk:
@birdstorm 
There is some comments here https://github.com/tikv/tikv/pull/10743, PTAL.

birdstorm:
> @birdstorm 
> There are some comments here https://github.com/tikv/tikv/pull/10743, PTAL.

@cfzjywxk LGTM.

proto/kvrpcpb.proto
birdstorm: After the users have upgraded to v5.0, we can update tispark version to v2.4 as well.

birdstorm (approved):
LGTM

proto/kvrpcpb.proto
cfzjywxk: @birdstorm 
Is it possible that there exists a time window in which tispark v3.0-hotfix reads the tikv-5.0 and could not parse the newly added boolean field from the `ScanResponse`?

birdstorm: > @birdstorm 
> Is it possible that there exists a time window in which tispark v3.0-hotfix reads the tikv-5.0 and could not parse the newly added boolean field from the `ScanResponse`? 

I think we could actually use the same ScanResponse structure as v5.0 tikv

disksing:
I don't quite understand. What is it used for?

disksing:
I think we can extract locks from key error.

ekexium:
> I think we can extract locks from key error.

Yes, it's what TiSpark currently does. But it is said that the scan of locks causes severe performance degradation. So an extra field can help avoid unnecessary scans. cc @birdstorm

disksing (approved):

PR #797, opened 2021-08-22
debugpb: add GetAllRegionsInStore()

andylokandy:


proto/debugpb.proto
```diff
@@ -248,3 +251,10 @@ message GetClusterInfoRequest {
 message GetClusterInfoResponse {
     uint64 cluster_id = 1;
 }
+
+message GetAllRegionsInStoreRequest {
+}
+
+message GetAllRegionsInStoreResponse {
+    repeated uint64 regions = 1;
```
disksing, line 259: Is id enough? I think we may add more information for debug purpose.

andylokandy: The `RegionInfo` rpc should be able to return the detailed information for each regon.

disksing: ok

disksing (approved):

Connor1996 (approved):
LGTM

PR #369, opened 2019-03-22
*: add get operator service

nolouch:
- Add get operator service.
- GetStoreResponse add status.

PR #800, opened 2021-08-26
Add `damaged_regions_id` for `StoreStats`

hunterlxt:
Add 3 operator modes for now


disksing:
> @disksing actually, it resued server.handler

I mean can you use HTTP API directly instead using gRPC?

hunterlxt:
> > @disksing actually, it resued server.handler
> 
> I mean can you use HTTP API directly instead using gRPC?

I'm worried about TLS etc. Unified into grpc may avoid some problems, we can configure the network(pd - tikv) uniformly

BusyJay (approved):
Giving that we have `GetOperator`, it seems reasonable to add an `AddOperator`.

HunDunDM:
How to ensure the compatibility of move-peer?
And I don't think `GetOperator` should exist either, but it is an ancient existence #369 and cannot be deleted now. It should not be a reason to continue to add new ones.

HunDunDM (commented):
lgtm

PR #806, opened 2021-09-02
BatchPut supports independent ttl for every key

daimashusheng:
my issue: https://github.com/pingcap/kvproto/issues/807

I add ttls param so that I can set independent ttl for every key in BatchPut request


disksing:
Hi @daimashusheng could you make the PR title more specific? And sign the CLA?

daimashusheng:
> Hi @daimashusheng could you make the PR title more specific? And sign the CLA?

I recreate the pull just to validate CLA

disksing:
@daimashusheng you will have to click the _CLA not signed yet_ button in the first comment.

andylokandy:
This is a breaking change. I think there should be a way to introduce TTLs while keeping ttl as before. Eg., provide a new ttls field that disallows ttl when set.

disksing:
> This is a breaking change. I think there should be a way to introduce TTLs while keeping ttl as before. Eg., provide a new ttls field that disallows ttl when set.

Actually, repeated field is compatible with non-repeated field in wire format. As long as we take good care of the case when there is only one item in the list.

tisonkun:
@daimashusheng  I'm highly curious about the background about this change without any description. Could you elaborate a bit more?

disksing:
hi @daimashusheng , if you need any help. Please contact me via i@disksing.com

tisonkun:
closed as invalid. @daimashusheng as @disksing suggested, you can have a conversation to get the context about this repository and expected patches. Also you can post a topic and discuss publicly at internals.tidb.io.

Thanks for you participant.

daimashusheng:
> Hi @daimashusheng could you make the PR title more specific? And sign the CLA?

Done

tisonkun:
@daimashusheng please elaborate about the background of this PR and you'd better associate it with an issue.

It is hard to know why you make some changes without any description.

tisonkun:
@marsishandsome please comment the reason before or after reopen the issue instead of silently change it. If you have an offline discussion with @daimashusheng , convey it here or just let @daimashusheng do the action.

sunxiaoguang:
> This is a breaking change. I think there should be a way to introduce TTLs while keeping ttl as before. Eg., provide a new ttls field that disallows ttl when set.

We can't simply break existing interface like this. It can silently corrupt data which is definitely not acceptable.

marsishandsome:
> @marsishandsome please comment the reason before or after reopen the issue instead of silently change it. If you have an offline discussion with @daimashusheng , convey it here or just let @daimashusheng do the action.

Already discussed with @daimashusheng, he will address the comments.

daimashusheng:
> This is a breaking change. I think there should be a way to introduce TTLs while keeping ttl as before. Eg., provide a new ttls field that disallows ttl when set.

thank you, I have already modifed code and committed it again

disksing:
Also /cc @andylokandy @tisonkun

daimashusheng:
> @sunxiaoguang @daimashusheng As I have mentioned protobuf supports changing a non-repeated field to repeated. Their wire formats are the same, except for that there can only be 1 item when the field is not repeated. As a result, as long as we take care the case when there is only 1 TTL in the request, it can remain compabible.

Do you mean that I just need to modify ttl repeated and judge whether its size is 1 when I use it?

disksing:
> > @sunxiaoguang @daimashusheng As I have mentioned protobuf supports changing a non-repeated field to repeated. Their wire formats are the same, except for that there can only be 1 item when the field is not repeated. As a result, as long as we take care the case when there is only 1 TTL in the request, it can remain compabible.
> 
> Do you mean that I just need to modify ttl repeated and judge whether its size is 1 when I use it?

Exactly!

sunxiaoguang:
> @sunxiaoguang @daimashusheng As I have mentioned protobuf supports changing a non-repeated field to repeated. Their wire formats are the same, except for that there can only be 1 item when the field is not repeated. As a result, as long as we take care the case when there is only 1 TTL in the request, it can remain compabible.

Hm, interesting. What does code generated with non-repeated field do when serialized data has more than 1 repeated items?

sunxiaoguang:
> > @sunxiaoguang @daimashusheng As I have mentioned protobuf supports changing a non-repeated field to repeated. Their wire formats are the same, except for that there can only be 1 item when the field is not repeated. As a result, as long as we take care the case when there is only 1 TTL in the request, it can remain compabible.
> 
> Hm, interesting. What does code generated with non-repeated field do when serialized data has more than 1 repeated items?

Found people discussing this question, looks like old code will see the last element for this case.

daimashusheng:
> @sunxiaoguang @daimashusheng As I have mentioned protobuf supports changing a non-repeated field to repeated. Their wire formats are the same, except for that there can only be 1 item when the field is not repeated. As a result, as long as we take care the case when there is only 1 TTL in the request, it can remain compabible.

Thank you, I have already updated and committed codes of kvproto and tikv,  and I have tested successfully for clients both reaptead and non-repeated ttl

disksing:
It should be fine as long as we don't use a newer client to write data to an old tikv.

disksing (approved):

proto/kvrpcpb.proto
```diff
@@ -459,7 +459,7 @@ message RawBatchPutRequest {
     Context context = 1;
     repeated KvPair pairs = 2;
     string cf = 3;
-    uint64 ttl = 4;
+    repeated uint64 ttl = 4;
```
andylokandy, line 462:
```suggestion
    repeated uint64 ttls = 4;
```

daimashusheng: it is not compatile with old version if it is modified to ttls, as you said before.

andylokandy: renaming in protobuf is compatible as long as the number remains the same.

I don't know that changing singular field to repeated will be compatible, and thanks @disksing for pointing out.

tisonkun: > renaming in protobuf is compatible as long as the number remains the same.

It is wire format compatible, but the generated code will change API names, and cause compile fail when upgrading.

tisonkun:
@marsishandsome you should sign off the CLA _for this repo_ also...

> [CLA assistant check](https://cla-assistant.io/pingcap/kvproto?pullRequest=806)
> Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you all sign our [Contributor License Agreement](https://cla-assistant.io/pingcap/kvproto?pullRequest=806) before we can accept your contribution.
> **1** out of **2** committers have signed the CLA.
> 
> ✅ daimashusheng
> ❌ marsishandsome
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/pingcap/kvproto?pullRequest=806) it.

proto/kvrpcpb.proto
andylokandy: > cause compile fail when upgrading 

This is intended. The new tikv server and client should re-implement the logic on TTL.

andylokandy (approved):

andylokandy:
@daimashusheng Thank you!

PR #814, opened 2021-09-22
resource_usage_agent: support summary data reporting

mornyx:
Signed-off-by: mornyx <mornyx.z@gmail.com>


disksing (approved):

proto/resource_usage_agent.proto
```diff
@@ -17,6 +17,10 @@ service ResourceUsageAgent {
     // Report the CPU time records. By default, the records with the same
     // resource group tag will be batched by minute.
     rpc ReportCPUTime(stream CPUTimeRecord) returns (EmptyResponse) {}
+
+    // Report the summary records. By default, the records with the same
+    // resource group tag will be batched by minute.
+    rpc ReportSummary(stream SummaryRecord) returns (EmptyResponse) {}
```
breeswish, line 23: Shall we integrate into one structure? (as they should share the same resource_group_tag)

proto/resource_usage_agent.proto
```diff
@@ -31,4 +35,14 @@ message CPUTimeRecord {
...
 }
 
+message SummaryRecord {
+    bytes resource_group_tag = 1;
+
+    // The number of reads of keys associated with resource_group_tag.
+    uint32 read_keys = 2;
+
+    // The number of writes of keys associated with resource_group_tag.
+    uint32 write_keys = 3;
```
breeswish, line 45: Do you plan to distinguish index keys and row keys?

mornyx: Can we use the label(used to distinguish row and index) as part of tipb.ResourceGroupTag? In this case, tikv does not need to perceive any difference between `row` and `index`, nor does it need to add additional fields. It only needs to transparently pass the ResourceGroupTag to the agent, and the agent will make a semantic distinction.

proto/resource_usage_agent.proto
mornyx: This is possible, but there are some structural inconsistencies between the cpu statistics and the summary statistics in tikv, and some integration work is needed.

proto/resource_usage_agent.proto
breeswish: Got it. This is a brilliant idea!

proto/resource_usage_agent.proto
breeswish: Could you explain a little bit about what kind of inconsistencies are there? It is not as simple as something I thought previously and I would like to have a deeper learning about the design.

mornyx: CPU statistics usually have a `time window` attribute, and the resource usage related to each tag is bound to the time period (usually one second).
However, the summary statistics are bound to the `request`, and there is no strict time window for the final statistics. (But according to the final demand, maybe the time window still needs to be bound to summary statistics in the end.)
Therefore, the two types of resource usage statistics are only temporarily distinguished in the whole process. If more appropriate, we can also integrate them.

breeswish: Maybe we have some misunderstandings about the scanned lines. According to my understanding, scanned lines are similar to CPU time, recorded for each resourceGroupTag for each second. In UI, when user selects a range of time, user sees the CPU time of each second in that range of time, and the scanned lines of each second in that range of time as well. /cc @crazycs520 Please help verify the behaviour.

mornyx: Oh, it looks like there is a mistake in my understanding. I will relearn the product form today and then make some adjustments to the code :)

proto/resource_usage_agent.proto
```diff
@@ -14,21 +14,27 @@ option java_package = "org.tikv.kvproto";
 
 // ResourceUsageAgent is the service for storing resource usage records.
 service ResourceUsageAgent {
-    // Report the CPU time records. By default, the records with the same
+    // Report the resource usage records. By default, the records with the same
     // resource group tag will be batched by minute.
-    rpc ReportCPUTime(stream CPUTimeRecord) returns (EmptyResponse) {}
+    rpc Report(stream ResourceUsageRecord) returns (EmptyResponse) {}
```
disksing, line 19: Change the RPC name should be a change brokes compatibility. Not sure if we really want it.

mornyx: > Change the RPC name should be a change brokes compatibility. Not sure if we really want it.

It seems that if the RPC interface is integrated, the method name needs to be changed (it better be), and there is a break change.
Please help confirm whether we can do this, or add a new method to make the transition. /cc @breeswish @zhongzc

breeswish: Yes, changing RPC name is a breaking change.

breeswish: How about leaving `ReportCPUTime`, but mark it as deprecated? Then we can add a new interface `Report`.

New clients should never call `ReportCPUTime` any more. But current receivers should accept both `ReportCPUTime` and `Report`.

mornyx: Fine, this is the most suitable way.

proto/resource_usage_agent.proto
```diff
@@ -14,9 +14,15 @@ option java_package = "org.tikv.kvproto";
 
 // ResourceUsageAgent is the service for storing resource usage records.
 service ResourceUsageAgent {
+    // DEPRECATED: We now use `Report` to report not only CPU time.
+    //
     // Report the CPU time records. By default, the records with the same
     // resource group tag will be batched by minute.
     rpc ReportCPUTime(stream CPUTimeRecord) returns (EmptyResponse) {}
```
breeswish, line 21: We can change CPUTimeRecord to ResourceUsageRecord, since this change is compatible.

breeswish (approved):
The rest LGTM

proto/resource_usage_agent.proto
mornyx: I’m not sure before, it seems that the type name of the parameter does not affect compatibility?

mornyx: My idea is to keep the original proto text unchanged to avoid the illusion of breaking compatibility, and these will be deleted when the transition is completed.

PR #819, opened 2021-09-29
br: Add option to HDFS export backend

peng1999:
Signed-off-by: Peng Guanwen <pg999w@outlook.com>

PR #820, opened 2021-10-11
Update go pkg

peng1999:
This PR generated the corresponding go code that #819 forgot.


disksing:
There are many inrelevant changes. I think it would be better to reset them.

peng1999:
> There are many inrelevant changes. I think it would be better to reset them.

Ok. All changes except `brpb.pb.go` has been reverted.

PR #815, opened 2021-09-23
Add a new key error type AssertionFailed

MyonKeminta:
This might be part of https://github.com/pingcap/tidb/issues/26833

PR #824, opened 2021-10-19
Update kvproto for optimizing assertion for pessimistic transactions

MyonKeminta:
This continues https://github.com/pingcap/kvproto/pull/815 , which is part of https://github.com/pingcap/tidb/issues/26833 .

Since there need to be some adjustments to the design, the kvproto still needs updating. And thus this PR is still not suggested too quickly...
I think it's not a problem to submit a PR that is not ready to merge yet, since the PR can be a tool for tracking progress.


proto/kvrpcpb.proto
```diff
@@ -949,6 +956,15 @@ enum Assertion {
      NotExist = 2;
 }
 
+enum AssertionLevel {
+    // No assertion.
+    Off = 0;
+    // Do assertion when it can be done without affecting performance.
```
ekexium, line 962:
```suggestion
    // Assertion is enabled, but not enforced when it might impact performance.
```

proto/kvrpcpb.proto
```diff
@@ -949,6 +956,15 @@ enum Assertion {
...
+    Fast = 1;
+    // Do full assertion even if the performance might be slown down.
```
ekexium, line 964:
```suggestion
    // Assertion is enabled and enforced.
```

ekexium (approved):

cfzjywxk (approved):
LGTM

PR #828, opened 2021-10-25
`GlobalConfig` Rule them all

LemonHX:
added `StoreGlobalConfig` and `LoadGlobalConfig` for pdpd as grpc methods.


proto/pdpb.proto
```diff
@@ -84,6 +84,39 @@ service PD {
...
+    string name = 1;
+    string value = 2;
+}
+
+message StoreGlobalConfigResponse {
+    Error error = 1;
+}
+
+message LoadGlobalConfigRequest {
+    repeated string name = 1;
```
crazycs520, line 114:
```suggestion
    repeated string names = 1;
```

proto/pdpb.proto
```diff
@@ -84,6 +84,39 @@ service PD {
...
+}
+
+message LoadGlobalConfigResponse {
+    repeated string value = 1;
```
crazycs520, line 118:
```suggestion
    repeated string values = 1;
```

proto/pdpb.proto
```diff
@@ -84,6 +84,39 @@ service PD {
...
+
+    rpc WatchGlobalConfig(WatchGlobalConfigRequest) returns (stream WatchGlobalConfigResponse) {}
+}
+
+message WatchGlobalConfigRequest {
+
+}
+
+message WatchGlobalConfigResponse {
+    string key = 1;
```
crazycs520, line 100:
```suggestion
    string name = 1;
```

proto/pdpb.proto
```diff
@@ -84,6 +84,39 @@ service PD {
...
+    string value = 2;
+}
+
+message StoreGlobalConfigRequest {
+    string name = 1;
+    string value = 2;
+}
+
+message StoreGlobalConfigResponse {
+    Error error = 1;
```
crazycs520, line 110: Look like this is field is never used?

crazycs520: Maybe you can just like `LoadGlobalConfigResponse`. just use `ok` filed.

proto/pdpb.proto
```diff
@@ -84,6 +84,39 @@ service PD {
...
+}
+
+message LoadGlobalConfigRequest {
+    repeated string names = 1;
+}
+
+message LoadGlobalConfigResponse {
+    repeated string values = 1;
+    bool ok = 2;
```
crazycs520, line 119: Maybe should use `Error` filed, just like `StoreGlobalConfigResponse`.

JmPotato (approved):

crazycs520 (approved):

PR #832, opened 2021-11-02
Return the lock TS of the unexpected lock in TxnNotFound

sticnarf:
ref: https://github.com/tikv/tikv/issues/11148

If TiKV returns the lock TS, the client can choose whether to rollback the transaction itself.


disksing (approved):

disksing:
can we merge it?

sticnarf:
> can we merge it?

Let's wait for review of https://github.com/tikv/tikv/pull/11208 first. If other people don't have objections on the solution, then we can merge.

sticnarf:
Close because we change the way of resolving the issue.

PR #836, opened 2021-11-22
br: add br backup info

YuJuncen:
cc https://github.com/pingcap/tidb/issues/29969

For better utilize the already defined `StorageBackend` and better encoding binary keys, I decided to encode the taskinfo as protocol buffer. 🤔


proto/brpb.proto
```diff
@@ -184,6 +184,36 @@ message BackupRequest {
...
+    repeated string table_filter = 5;
+
+    // The last timestamp of the task has been updated. 
+    // This is a simple solution for unfrequent config changing: 
+    // When we watched a config change(via polling or etcd watching),
+    // We perform a incremental scan between [last_update_ts, now), 
+    // for filling the diff data during conf changing.
+    // The current implementation scan [0, now) for every task ranges newly added, 
+    // So this field is reserved for future usage.
+    reserved 6; reserved "last_backup_ts";
```
3pointer, line 208: `last_backup_ts` => `last_update_ts` 🤔
 
shall we have something like `create ts`?

proto/brpb.proto
```diff
@@ -184,6 +184,36 @@ message BackupRequest {
     CipherInfo cipher_info = 14;
 }
 
+message StreamBackupTaskInfo {
+    // The storage for backup, parsed by BR.
+    StorageBackend storage = 1;
+    // The time range for backing up.
+    uint64 start_ts = 2;
+    uint64 end_ts = 3;
```
3pointer, line 192: will tikv remove this task if `resolved_ts` reached `end_ts`?

proto/brpb.proto
YuJuncen: I guess the `start-ts` would perform as create ts?

proto/brpb.proto
YuJuncen: I guess we can leave it a "done" status and let the user to delete?

3pointer (approved):
LGTM

proto/brpb.proto
3pointer: `start-ts` could be earlier than task create time. but for now it's not required.

kennytm:
undo the changes to `pkg/coprocessor/coprocessor.pb.go` and `pkg/pdpb/pdpb.pb.go`?

proto/brpb.proto
```diff
@@ -184,6 +184,41 @@ message BackupRequest {
...
+    // for filling the diff data during conf changing.
+    // The current implementation scan [0, now) for every task ranges newly added, 
+    // So this field is reserved for future usage.
+    reserved "last_update_ts";
+
+    // The prefix at etcd (or other metadata storage) for finding the ranges of the task.
+    // We use '/tidb/br-stream/ranges/<task-name>/<start-key> -> <end-key>' 
+    // as the default key prefix for the target ranges for the task.
+    // Given the path unchanged, the dynamic config(or, hyperlink?) seems useless.
+    reserved "backup_range_key_prefix";
```
kennytm, line 214: why do we even need this config 🤔

kennytm (commented):
rest lgtm

proto/brpb.proto
YuJuncen: Just a placeholder for the dead 'ranges' 🤔, maybe just remove it

YuJuncen: Rename it into "ranges" in memory of the original `ranges` field.

proto/brpb.proto
joccau: The Task would be exist, but the observing of stream-log has been stoped. And user can status/stop the task in this situation ?

disksing (approved):

PR #838, opened 2021-12-06
br: add azure message

Leavrth:


YuJuncen (approved):

YuJuncen:
LGTM, BTW I guess maybe register `Azure` at external storage too?
image

proto/brpb.proto
```diff
@@ -236,6 +237,27 @@ message GCS {
...
+    string storage_class = 4;
+
+    string account_name = 5;
+    // Use shared key to access the azure blob
+    // If empty and use_bearer_token is false, read shared key in the environment.
+    string shared_key = 6;
+    // Prefer to use token to access the azure blob
+    // Need variables in environment: $AZURE_CLIENT_ID, $AZURE_TENANT_ID, $AZURE_CLIENT_SECRET
+    // See https://docs.microsoft.com/en-us/azure/storage/common/identity-library-acquire-token?toc=/azure/storage/blobs/toc.json
+    bool use_bearer_token = 7;
```
kennytm, line 258: 🤔 how do we define this flag? i think it's better for TiKV to implicitly determine from their nodes' environment.

Leavrth: Done. Please take another look.

kennytm (approved):

proto/brpb.proto
```diff
@@ -236,6 +237,29 @@ message GCS {
...
 }
 
+// AzureBlobStorage storage backend saves files into azure blob storage.
+message AzureBlobStorage {
+    string endpoint = 1;
+    // Alias: container
+    string bucket = 2;
+    // Notice: prefix starts without `/`, otherwise the first directory's name is empty.
+    string prefix = 3;
+    // Alias: access_tier. If empty, set `Hot` as default
```
kennytm, line 247: oh i thought we have already agreed that the default access_tier is to be decided later.

```suggestion
    // Alias: access_tier.
```

PR #844, opened 2021-12-23
fix RawBatchPut api forward compatibility

iosmanthus:
For now, the `ttls`'s field number is reused the `ttl`'s which will introduce a forward compatibility issue: https://github.com/tikv/client-java/issues/419.

This issue will cause the master client-java including client-go can't access TiKV by calling `RawBatchPutRequest`. This pull request should control the radius of influence only covering TiKV v5.3.0. However, client-java with this commit doesn't break while accessing the TiKV v5.3.0 calling `RawBatchPutRequest` which is conflicted with the promise of protobuf.


iosmanthus:
@BusyJay PTAL

disksing (approved):

PR #846, opened 2021-12-28
Add tracing

andylokandy:
Signed-off-by: andylokandy <andylokandy@hotmail.com>


zhongzc (approved):

proto/trace.proto
```diff
@@ -0,0 +1,43 @@
...
+message Property {
+    string key = 1;
+    string value = 2;
+}
+
+service TraceRecordPubSub {
+    // Clients subscribe tracing records through this RPC, the subscribed component will periodically (e.g. per minute)
+    // publishes tracing records to clients via gRPC stream.
+    rpc Subscribe(TraceRecordRequest) returns (stream TraceRecord) {}
+}
```
breeswish, line 41: I have some worries for this part, as it may be not compatible with how TiDB actually use. The TiDB will expect to receive trace results through the response synchronously.

proto/trace.proto
```diff
@@ -0,0 +1,43 @@
...
+message Span {
+    // The unique span id within a `TraceRecord`.
+    uint32 span_id = 1;
+    // The parent span within a `TraceRecord`.
+    // Set to 0 to indicate that it's the root span within a `TraceRecord`.
+    uint32 parent_id = 2;
+    uint64 begin_unix_ns = 3;
+    uint64 duration_ns = 4;
+    string event = 5;
+    repeated Property properties = 6;
```
breeswish, line 29: I do have some worries about the encoding cost when you use string as event names.

andylokandy: I think it should be pretty easy to implement a stream-wise event dictionary later.

proto/trace.proto
andylokandy: Yes, all reporting will be asynchronous in this design.

marsishandsome (approved):
LGTM

marsishandsome:
Is there any public design documents like RFC?

breeswish (approved):

disksing:
@andylokandy seems you lost some file.

andylokandy:
@disksing Thanks. I have submitted the files.

breeswish (approved):

proto/trace.proto
```diff
@@ -19,8 +19,14 @@ message TraceContext {
     uint64 parent_id = 2;
 }
 
-// The tracing spans for a request on a service.
 message TraceRecord {
+    oneof record_oneof {
+        SpanSet spans = 1;
```
breeswish, line 23: What will it be, if it is not a span?

andylokandy: It could be the event name dictionary.

disksing (approved):

PR #849, opened 2021-12-30
kvproto: add buckets related definitions --- ref #11759

tonyxuqqi:
Ref Tikv #11759.
This is part of the code to address work item Tikv #11759.


proto/pdpb.proto
```diff
@@ -797,3 +799,16 @@ enum QueryKind {
...
 }
+
+message RegionStatsRequest {
+    RequestHeader header = 1;
+    // Peer sending the stats--it can be leader or follower (in stale read case).
+    metapb.Peer peer = 2;
+    metapb.Buckets buckets = 3;
+}
+
+message RegionStatsResponse {
```
BusyJay, line 810: What's the purpose of the response?

proto/metapb.proto
```diff
@@ -61,6 +61,16 @@ message RegionEpoch {
     uint64 version = 2;
 }
 
+message Buckets {
+    uint64 region_id = 1;
+    uint64 version = 2; // A hint indicate if keys have changed.
+    repeated bytes keys = 3;
+    repeated uint64 read = 4;
+    repeated uint64 scan = 5;
+    repeated uint64 write = 6;
```
5kbpers, line 70: How about using [QueryStats](https://github.com/pingcap/kvproto/blob/master/proto/pdpb.proto#L772) here?

proto/metapb.proto
```diff
@@ -73,6 +83,7 @@ message Region {
     // The field is only used by PD and should be ignored otherwise.
     // If encryption_meta is empty (i.e. nil), it means start_key and end_key are unencrypted.
     encryptionpb.EncryptionMeta encryption_meta = 6;
+    Buckets buckets = 7;
```
BusyJay, line 86: I was thinking the buckets are queries and updated in a standalone background fashion. For example, client can use stale buckets to split queries without actually depending on the information from Region.

proto/metapb.proto
```diff
@@ -61,6 +61,16 @@ message RegionEpoch {
     uint64 version = 2;
 }
 
+message Buckets {
```
BusyJay, line 64: How about adding some comments?

proto/metapb.proto
```diff
@@ -61,6 +61,16 @@ message RegionEpoch {
...
+    uint64 region_id = 1;
+    uint64 version = 2; // A hint indicate if keys have changed.
+    repeated bytes keys = 3;
+    repeated uint64 read = 4;
+    repeated uint64 scan = 5;
+    repeated uint64 write = 6;
+    uint64 period_in_us = 7; // The period in microseconds that stats are collected with in.
```
BusyJay, line 71: ms should be enough.

proto/metapb.proto
BusyJay: The related part of RFC is in https://github.com/tikv/rfcs/blob/8bd15f24fd01488a6d639d5a13e7876ddb3159da/text/0082-dynamic-size-region.md?plain=1#L104-L111.

proto/pdpb.proto
```diff
@@ -797,3 +799,19 @@ enum QueryKind {
     Commit = 10;
     Rollback = 11;
 }
+
+message RegionStatsRequest {
+    RequestHeader header = 1;
+    // Peer sending the stats--it can be leader or follower (in stale read case).
+    metapb.Peer peer = 2;
+    metapb.Buckets buckets = 3;
+    uint64 read = 4;
```
bufferflies, line 808: Does the read means the total read of the region?

proto/metapb.proto
tonyxuqqi: Is there any problem of using existing GetRegion API to get the latest version of buckets? Client can still have the logic described in the RFC. 
To me the region's other fields are very minimal. Having a specific buckets API is not necessary.

proto/metapb.proto
tonyxuqqi: ok

proto/metapb.proto
tonyxuqqi: Today pdpb.proto depends on this file.  This file cannot import pdpb.proto or it will leads to cycle import.

proto/metapb.proto
tonyxuqqi: done

proto/metapb.proto
tonyxuqqi: For reporting bucket stats, we will use stream API which will be defined in Minhua's PR

proto/metapb.proto
tonyxuqqi: But I think it's a good idea to wrap all stats data into a separate BucketStats message, which will be easier to extend in the future.

proto/metapb.proto
```diff
@@ -61,6 +61,33 @@ message RegionEpoch {
...
+    uint64 region_id = 1;
+
+    // A hint indicate if keys have changed.
+    uint64 version = 2;
+
+    // keys of buckets, not include start/end key of region
+    repeated bytes keys = 3;
+
+    // bucket stats
+    repeated BucketStats stats = 4;
```
BusyJay, line 85: It's a bad idea. And it's explained in the RFC PR why the stats are flatten: https://github.com/tikv/rfcs/pull/82#discussion_r764879850

proto/metapb.proto
BusyJay: There are 3 reasons:
1. buckets change frequently, they don't need to be persisted, while region must be persisted;
2. if buckets are associated with stats, then we don't need to do anything specific for matching buckets with stats;
3. unlike regions, buckets are not available all the time. For example, if a peer is elected as a new leader, or split worker is busy, or the node is restarted, then buckets may not be available for quite a long time.

Given the difference between buckets and regions, handle them respectively is more flexible and easy to implement correctly.

proto/metapb.proto
tonyxuqqi: Yeah, actually the repeated can be removed here. Then it's OK.

BusyJay (commented):
The query of buckets are still missing, rest LGTM

BusyJay (approved):

proto/pdpb.proto
```diff
@@ -796,4 +796,4 @@ enum QueryKind {
     AcquirePessimisticLock = 9;
     Commit = 10;
     Rollback = 11;
-}
+}
```
5kbpers, line 799:
```suggestion
}

```

5kbpers (commented):
rest LGTM

proto/metapb.proto
```diff
@@ -61,6 +61,42 @@ message RegionEpoch {
     uint64 version = 2;
 }
 
+message BucketStats {
```
lhy1024, line 64: We will no longer count keys? Also, previously in region and store, queries were separable by type, but now only get and scan can be distinguished?

lhy1024 (commented):
I think we may need to refer to QueryStats

proto/metapb.proto
```diff
@@ -61,6 +61,42 @@ message RegionEpoch {
...
+    repeated uint64 scan_bytes = 2;
+
+    // total write in bytes of each bucket
+    repeated uint64 write_bytes = 3;
+
+    // total read (point get) qps of each bucket
+    repeated uint64 read_qps = 4;
+
+    // total scan qps of each bucket
+    repeated uint64 scan_qps = 5;
```
nolouch, line 78: Is it necessary to distinguish between point and scan？I think the first version can be simpler.

tonyxuqqi: ok

5kbpers (approved):

lhy1024 (approved):

PR #853, opened 2022-01-17
Add find grpc and protobuf in cmake, to co-work with bundle grpc for tiflash

zanmato1984:
Add find grpc and protobuf in cmake, so that kvproto can be built both individually and bundled within tiflash/client-c.


JaySon-Huang (approved):
LGTM

solotzg:
TODO: cherry-pick it to each 5.x branch @zanmato1984

PR #855, opened 2022-01-19
br-stream: add apply interface for restore

3pointer:


proto/import_sstpb.proto
```diff
@@ -280,3 +283,34 @@ message DuplicateDetectResponse {
...
+    backup.CipherInfo cipher_info = 6;
+}
+
+message ApplyResponse {
+    // The actual key range (after rewrite) of the downloaded file. The range is
+    // inclusive in both ends.
+    Range range = 1 [(gogoproto.nullable) = false];
+
+    Error error = 2;
+}
```
disksing, line 316: Please add an extra empty line at EOF.

proto/import_sstpb.proto
```diff
@@ -280,3 +283,34 @@ message DuplicateDetectResponse {
     // ]
     repeated KvPair pairs = 3;
 }
+
+message ApplyRequest {
+    // The file name of the SST file.
+    string name = 2;
```
disksing, line 289: why does it start from 2?

disksing: and no 5?

3pointer: just like `DownloadRequest` reserved some number for future. I think maybe it's some traditions...🤔

kennytm: @3pointer no, numbers are skipped only because they were used in fields in older versions which were then removed. It's ok to give the tags 1, 2, 3, 4 here.

3pointer: Ok, fixed.

YuJuncen (approved):

disksing (approved):

PR #859, opened 2022-01-24
Add apply busy field and update raft 

Connor1996:
Used for https://github.com/tikv/tikv/pull/11409


Cargo.toml
```diff
@@ -20,7 +20,7 @@ prost = { version = "0.7", optional = true }
 prost-derive = { version = "0.7", optional = true }
 futures = "0.3.5"
 grpcio = { version = "0.9.0", default-features = false, features = ["secure"] }
-raft-proto = { version = "0.6.0-alpha", default-features = false }
+raft-proto = { version = "0.7.0", default-features = false }
```
BusyJay, line 23: It's not published yet.

BusyJay (approved):
Rest LGTM

proto/pdpb.proto
```diff
@@ -573,6 +573,8 @@ message StoreStats {
     uint64 slow_score = 22;
     // Damaged regions on the store that need to be removed by PD.
     repeated uint64 damaged_regions_id = 23;
+    // If the apply worker is busy, namely high apply wait duration
+    bool is_apply_busy = 24;
```
NingLin-P, line 577: Is this flag used for https://github.com/tikv/tikv/pull/11859? Store heartbeat has an interval of 10s, when PD detects such apply busy store and make the scheduling, the latency spike may already happen. As https://github.com/tikv/tikv/pull/11859#issuecomment-1014476371, the issue may simply be fixed by also checking the size of lagging apply/replicate logs instead of just the number. /cc @BusyJay

Connor1996: Yes. We can let raftstore trigger store heartbeat immediately when apply busy regardless of the interval of 10s. 

We are planning to move the reject transfer logic to the PD side:
1. when uptime < 2s, reject to consider the case that logs are in fetching, so apply is not busy now
2. when 2s < uptime < 60s and apply is busy, reject
3. uptime > 60s, do not reject anymore

NingLin-P (approved):

PR #863, opened 2022-02-15
Add recovery in progress error

Connor1996:
When doing online recovery, we'd like the region to reject writes with the error `RecoveryInProgress`.


disksing (approved):

proto/errorpb.proto
```diff
@@ -128,6 +128,11 @@ message DataIsNotReady {
     uint64 safe_ts = 3;
 }
 
+message RecoveryInProgress {
+    // The requested region ID
+    uint64 region_id = 1;
```
BusyJay, line 133: Isn't it part of the request?

Connor1996: nope, the request has the context, but the response doesn't.

BusyJay: What I mean client should know the request corresponding to the response, so it should know the region id.

Connor1996: Yes, but seems the error `NotLeader`, `RegionNotFound` and etc. includes the region id as well

BusyJay (approved):

proto/errorpb.proto
```diff
@@ -148,4 +153,6 @@ message Error {
     DataIsNotReady data_is_not_ready = 13;
     RegionNotInitialized region_not_initialized = 14;
     DiskFull disk_full = 15;
+    // Online recovery is still in performing, reject writes to avoid potential issues
+    RecoveryInProgress RecoveryInProgress = 16;
```
overvenus, line 157: Why not use snake_case? For the sake of consistency 😅

Connor1996: oh, miswrite

PR #865, opened 2022-02-16
pdpb: support getting region with buckets

youjiali1995:
Signed-off-by: youjiali1995 <zlwgx1023@gmail.com>


proto/pdpb.proto
```diff
@@ -258,6 +258,7 @@ message GetRegionResponse {
     // Pending peers are the peers that the leader can't consider as
     // working followers.
     repeated metapb.Peer pending_peers = 6;
+    metapb.Buckets buckets = 7;
```
youjiali1995, line 261: I can't find a frequent use case that doesn't need buckets, so always fetching it for simplicity.

BusyJay: What if the range doesn't change at all? It was designed to be updated periodically in the background. Otherwise the response may be too large to be processed quickly. Fetching a region with full bucket info can be the same as fetching 10GiB / 128MiB = 100 regions in the past given the size difference.

youjiali1995: > It was designed to be updated periodically in the background.

No, updated when the bucket's version is changed. TiDB can send requests with bucket version to TiKV and TiKV indicates TiDB to update buckets if the version isn't matched with its.

BusyJay: What if conf version is changed but bucket version is the same?

youjiali1995: Conf version mismatch doesn't return epoch-not-match error to TiDB, so the region in region cache won't be invalidated. If the version is changed, region and buckets should be both updated.

Buckets here may be wasted if the region version doesn't change but region is invalidated. It seldom happens in TiDB only when meets no-leader, region-not-found, key-not-in-region, store-not-match and other unknown errors.

I just realized that TiFlash may not need buckets information, so we need to provide a standalone way to get buckets of a region, i.e., the `GetBucket()` API.

BusyJay (approved):

PR #866, opened 2022-02-23
pdpb: support to report bucket

5kbpers:
Report buckets and bucket stats to pd


BusyJay (approved):

proto/pdpb.proto
```diff
@@ -799,3 +801,13 @@ enum QueryKind {
...
 }
+
+message ReportBucketsRequest {
+    RequestHeader header = 1;
+
+    metapb.RegionEpoch region_epoch = 2;
+    metapb.Buckets buckets = 3;
+}
+
+message ReportBucketsResponse {
```
bufferflies, line 812: how about carry RequestHeader?

5kbpers: What is it used for?

youjiali1995: @bufferflies Do you mean `ResponseHeader`? It's a convention to carry potential error?

bufferflies: yes, like cluster id is not right?

bufferflies (approved):

PR #867, opened 2022-02-24
pdpb: synchronizing region should support bucket.

bufferflies:
Signed-off-by: bufferflies <1045931706@qq.com>
synchronizing region should support bucket.


proto/pdpb.proto
```diff
@@ -448,6 +448,9 @@ message RegionHeartbeatResponse {
     //       1. replacing peers
     //       2. demoting voter directly
     ChangePeerV2 change_peer_v2 = 9;
+
+    // PD should contain the buckets information if origin region's epoch is changed.
+    metapb.Buckets buckets=10;
```
BusyJay, line 453: What's the purpose? And it should be formatted.

disksing:
@bufferflies any update?

proto/pdpb.proto
bufferflies: the operator should have buckets information like transterLeader.

bufferflies:
PTAL @BusyJay again

BusyJay (approved):

PR #868, opened 2022-02-28
metapb: refine store state

rleungx:
This PR introduces `NodeState` to replace `StoreState`, which is confusing sometimes, in the future.


pkg/brpb/brpb.pb.go
```diff
@@ -61,7 +61,7 @@ func (x CompressionType) String() string {
 	return proto.EnumName(CompressionType_name, int32(x))
 }
 func (CompressionType) EnumDescriptor() ([]byte, []int) {
-	return fileDescriptor_brpb_a90a70ff0dc32c8e, []int{0}
+	return fileDescriptor_brpb_7c3d2f8143177a80, []int{0}
```
disksing, line 64: would you like to reset the unrelated files?

rleungx: These differences are generated by running `make go`.

disksing: I know. You can checkout unrelated files to make the diff smaller.

rleungx: got it.

disksing (approved):

proto/metapb.proto
```diff
@@ -52,6 +63,8 @@ message Store {
     int64 last_heartbeat = 11;
     // If the store is physically destroyed, which means it can never up again.
     bool physically_destroyed = 12;
+    // NodeState is used to replace StoreState which will be deprecated in the future.
```
BusyJay, line 66: I think you can mark it as deprecated. See https://developers.google.com/protocol-buffers/docs/proto3#options.

proto/metapb.proto
```diff
@@ -26,6 +26,17 @@ enum StoreState {
     Tombstone = 2;
 }
 
+// NodeState is going to replace StoreState to make the state concept more clear.
+// "Up" is devided into "Preparing" and "Serving" stages so that we can better describe the online process.
+// "Removing" is just like previous `Offline` which is more accurate.
+// "Removed" has the same meaning with `Tombstone`.
+enum NodeState {
+    Preparing = 0;
```
BusyJay, line 34: How about adding docs to explaining each state?

BusyJay (commented):
Rest LGTM.

proto/metapb.proto
rleungx: Yes, once the related feature is finished, I will mark the old state as deprecated.

PR #870, opened 2022-03-01
*: move ResetToVersion to debugpb

longfangsong:
Signed-off-by: longfangsong <longfangsong@icloud.com>


nolouch (approved):

lhy1024 (commented):
so much file diff？

proto/debugpb.proto
```diff
@@ -259,3 +263,10 @@ message GetAllRegionsInStoreRequest {
 message GetAllRegionsInStoreResponse {
     repeated uint64 regions = 1;
 }
+
+message ResetToVersionRequest {
+    uint64 ts = 1;
+}
+
+message ResetToVersionResponse {
+}
```
disksing, line 272: Pleaase add an empty line here.

longfangsong:
> so much file diff？

I chaned `kvrpcpb` and IMO all protos which import it need to regenerate go file when `make go`.

proto/debugpb.proto
longfangsong: > Do we need a resp head here?

In fact, there're no infomation we can return here.

disksing (approved):

PR #871, opened 2022-03-02
coprocessor: Response contains latest buckets version

youjiali1995:
Copr requests will contains buckets version in the context and TiKV should return the up-to-date buckets information in the copr response if the version is stale to notify clients.


proto/coprocessor.proto
```diff
@@ -58,6 +58,9 @@ message Response {
     bool can_be_cached = 9;
 
     reserved 10;
+
+    // Contains the up-to-date buckets of the region if Request.context.buckets_version is stale.
+    metapb.Buckets buckets = 12;
```
BusyJay, line 63: I'm afraid this can be a large amount of data. How about making it a bool and let TiDB query PD instead giving that it's OK to have stale buckets.

youjiali1995: OK. To avoid TiDB querying PD one time each request, I will make it a uint64 buckets version and only when the buckets version in region cache is stale, TiDB queries PD.

youjiali1995: I removed the buckets version in the `Context` too. Now TiKV always returns the latest bucket version and TiDB updates buckets if its is stale.

BusyJay (approved):

Issue #874, opened 2022-03-03
Fail to make due to sed's error


buchuitoudegou:
When I run `make`, I got the following error:
```
...
sed：无法读取 s/import _ \"gogoproto\"//g：No such file or directory
sed：无法读取 s/import fmt \"fmt\"//g：No such file or directory
sed：无法读取 s/import io \"io\"//g：No such file or directory
sed：无法读取 s/import math \"math\"//g：No such file or directory
sed：无法读取 s/import _ \".*rustproto\"//：No such file or directory
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

sed：无法读取 s/import _ \"gogoproto\"//g：No such file or directory
sed：无法读取 s/import fmt \"fmt\"//g：No such file or directory
sed：无法读取 s/import io \"io\"//g：No such file or directory
sed：无法读取 s/import math \"math\"//g：No such file or directory
sed：无法读取 s/import _ \".*rustproto\"//：No such file or directory
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

sed：无法读取 s/import _ \"gogoproto\"//g：No such file or directory
sed：无法读取 s/import fmt \"fmt\"//g：No such file or directory
sed：无法读取 s/import io \"io\"//g：No such file or directory
sed：无法读取 s/import math \"math\"//g：No such file or directory
sed：无法读取 s/import _ \".*rustproto\"//：No such file or directory
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.

sed：无法读取 s/import _ \"gogoproto\"//g：No such file or directory
sed：无法读取 s/import fmt \"fmt\"//g：No such file or directory
sed：无法读取 s/import io \"io\"//g：No such file or directory
sed：无法读取 s/import math \"math\"//g：No such file or directory
sed：无法读取 s/import _ \".*rustproto\"//：No such file or directory
GO111MODULE=on go mod tidy
go: finding module for package github.com/pingcap/kvproto/pkg/rustproto
github.com/pingcap/kvproto/pkg/brpb imports
[...]

PR #875, opened 2022-03-03
scripts: improve sed version detect

disksing:

Fix https://github.com/pingcap/kvproto/issues/874


buchuitoudegou (approved):
LGTM

PR #879, opened 2022-03-09
import_sst: added `errorpb.Error` variant for the Error

YuJuncen:
We need the information of store for better retrying in the log restore (for the `Apply` RPC).


joccau (approved):
LGTM

3pointer (approved):

PR #880, opened 2022-03-10
add bucket swich for tikv

bufferflies:
Signed-off-by: bufferflies <1045931706@qq.com>


bufferflies:
@youjiali1995 @disking

youjiali1995 (commented):
I can't get it... Can you elaborate more on it?

BusyJay:
I thought the conclusion was to keep the current behavior that use configurations on both PD and TIKV to decide whether buckets are enabled for this sprint.

bufferflies:
> I can't get it... Can you elaborate more on it?

PD will notice tikv by PutStore and HeartBeat rpc.

bufferflies:
> I thought the conclusion was to keep the current behavior that use configurations on both PD and TIKV to decide whether buckets are enabled for this sprint.

In past, pd can depence on the buckets stream is open to decide the buckets behavior.

bufferflies:
it will use tikv config continue.

PR #888, opened 2022-03-28
br: support setting api versions in backup request

zz-jason:
Signed-off-by: Jian Zhang <zjsariel@gmail.com>


proto/brpb.proto
```diff
@@ -190,6 +190,14 @@ message BackupRequest {
...
     // The cipher_info is Used to encrypt sst
     CipherInfo cipher_info = 14;
+
+    // dst_api_version indicates the key-value encoding version used by the
+    // generated SST file. Accepted values:
+    // 1. "v1": the generated SST files are encoded with api-v1, can be restored
+    //    to clusters under v6.0.
+    // 2. "v2": the generated SST files are encoded with api-v2, can be restored
+    //    to clusrters >= v6.0.0 with api v2 enabled.
+    string dst_api_version = 15;
```
pingyu, line 200: Suggest to reuse [`kvrpcpbΩ.APIVersion`](https://github.com/pingcap/kvproto/blob/bd679cab506b8b8b75af2595624c3c3cd4254354/proto/kvrpcpb.proto#L723). Also for being consistent with [`api_version` in meta](https://github.com/pingcap/kvproto/blob/bd679cab506b8b8b75af2595624c3c3cd4254354/proto/brpb.proto#L72).

Suggest to add comments that this field is used for backup & restore cross TiKV clusters with different api versions.

proto/brpb.proto
```diff
@@ -190,6 +190,14 @@ message BackupRequest {
     int32 compression_level = 13;
     // The cipher_info is Used to encrypt sst
     CipherInfo cipher_info = 14;
+
+    // dst_api_version indicates the key-value encoding version used by the
+    // generated SST file. Accepted values:
+    // 1. "v1": the generated SST files are encoded with api-v1, can be restored
```
pingyu, line 196: "v1" can still be used on clusters above "v6.0". And the first version supporting "v2" would be "v6.1". Maybe it would be better to not mention the version.

proto/brpb.proto
disksing: what does `dst` mean?

zz-jason: @disksing `dst` is short for `destination`.

disksing (approved):

proto/brpb.proto
zz-jason: done

pingyu (approved):
LGTM

PR #896, opened 2022-04-08
added error message for log backup

YuJuncen:
The basic for error handing in log backup.

Signed-off-by: Yu Juncen <yujuncen@pingcap.com>


proto/brpb.proto
```diff
@@ -412,3 +412,12 @@ message DataFileInfo {
     // It may support encrypting at future.
     reserved "iv";
 }
+
+message LogBackupError {
```
3pointer, line 416: I preferred unified the name. LogBackup or StreamBackup

proto/brpb.proto
```diff
@@ -412,3 +412,12 @@ message DataFileInfo {
...
+    // the unix epoch time (in millisecs) of the time the error reported.
+    uint64 happen_at = 1;
+    // the unified error code of the error.
+    string error_code = 2;
+    // the user-friendly error message.
+    string error_message = 3;
```
joccau, line 422: error_message contains store_id and error reason ?

YuJuncen: store id should already be in the key path, anyway we can add it here.

3pointer (approved):

joccau (approved):
LGTM

PR #902, opened 2022-04-19
add prefix for clear request

3pointer:
in https://github.com/pingcap/tidb/issues/34098#issuecomment-1102293676. we think add a prefix is more safety.


YuJuncen (approved):

joccau (approved):
LGTM

PR #903, opened 2022-04-20
add ts for RawWriteBatch to be compatiable with API v2

iosmanthus:
Signed-off-by: iosmanthus <myosmanthustree@gmail.com>


proto/import_sstpb.proto
```diff
@@ -234,6 +234,7 @@ message WriteResponse {
 message RawWriteBatch {
     uint64 ttl = 1;
     repeated Pair pairs = 2;
+    uint64 ts = 3;
```
pingyu, line 237: All pairs have the same timestamp ?

pingyu: Where does `ts` come from ?
If `ts` is specified by request, the ingest data would be covered by earlier PUTs, and can not be CDC.
Another alternative is generated by TiKV.

iosmanthus: The `ts` should be generated by PD's TSO. `RawWriteBatch` would be sent to all the peers of a region concurrently. If `ts` is generated by TiKV, the batches would have different `ts` underlay.

pingyu: Then what will happen if underlaying `ts` are different ?

iosmanthus: If the timestamp is different, from the perspective of MVCC, a `ts` can read different versions of keys from different TiKV, especially during leader change.

pingyu: But RawKV gets do not provide a snapshot of a given `ts`. It always return the latest version.

pingyu (approved):
LGTM~

haojinming:
LGTM~

haojinming (approved):

PR #908, opened 2022-04-25
pd: RawKV GC SafePoint API

AmoebaProtozoa:
As described by https://github.com/tikv/rfcs/pull/90

Added 5 new RPC services as well as their corresponding messages related to RawKV GC


pingyu (approved):
LGTM~

pingyu:
@nolouch @MyonKeminta @TonsnakeLin PTAL, thanks ~

proto/pdpb.proto
```diff
@@ -73,6 +73,16 @@ service PD {
 
     rpc UpdateServiceGCSafePoint(UpdateServiceGCSafePointRequest) returns (UpdateServiceGCSafePointResponse) {}
 
+    rpc GetAllServiceGroups(GetAllServiceGroupsRequest) returns (GetAllServiceGroupsResponse) {}
```
nolouch, line 76: Can we use standalone like `gcpb.proto` ?

AmoebaProtozoa: do you mean both the `message` and `rpc` or just the `message`?
and should we also move old GC interface to a separate file?

AmoebaProtozoa: It's hard to separate out `gc` related messages as they depends on `RequestHeader` and `ResponseHeader`

proto/pdpb.proto
```diff
@@ -67,12 +67,6 @@ service PD {
 
     rpc ScatterRegion(ScatterRegionRequest) returns (ScatterRegionResponse) {}
 
-    rpc GetGCSafePoint(GetGCSafePointRequest) returns (GetGCSafePointResponse) {}
```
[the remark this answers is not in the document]
AmoebaProtozoa: Ok, So how about keeping the old GC apis in pdpb (until we decide to phase it out), adding only the new ones to a separate gcpb?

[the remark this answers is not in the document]
nolouch: yes.

proto/gcpb.proto
```diff
@@ -0,0 +1,100 @@
...
+
+  rpc UpdateGCSafePointByServiceGroup(UpdateGCSafePointByServiceGroupRequest) returns (UpdateGCSafePointByServiceGroupResponse) {}
+
+  rpc UpdateServiceSafePointByServiceGroup(UpdateServiceSafePointByServiceGroupRequest) returns (UpdateServiceSafePointByServiceGroupResponse) {}
+
+  rpc GetAllServiceGroupGCSafePoints(GetAllServiceGroupGCSafePointsRequest) returns (GetAllServiceGroupGCSafePointsResponse) {}
+}
+
+message GetAllServiceGroupsRequest {
+  pdpb.RequestHeader header = 1;
```
pingyu, line 29: It would better to be independent to PD's `RequestHeader` & `ResponseHeader`:
1. To make GC service possible to be totally independent to PD (e.g, a standalone micro service in Cloud).
2. Define error type of PD & GC service together seems not reasonable.

AmoebaProtozoa: Should the same separation be done on `PD` side? 
Say, move gc methods outside `grpc_service.go`? And use separate `type GcServer` to provide gc functionalities instead of `GrpcServer`? https://github.com/tikv/pd/blob/6ab2e4448c331de6dbedb61dbec5113e1a2a4122/server/grpc_service.go#L72
What about client? Is a separate `gc client` needed along side PD client?

pingyu: I think yes, and we can try. Unless it is too expensive to do so. @nolouch How do you think ?

nolouch: LGTM, PD client may include `gc client`?

pingyu:
LGTM~

nolouch (approved):

PR #911, opened 2022-04-28
add reject logappend extra msg type

tier-cap:
add reject logappend extra msg type to tell leader the log rejected.


disksing (approved):

v01dstar:
This newly added field has not been used, has it? I suggest not merging such proto change until the related PRs are ready, otherwise, it breaks others build, whoever is also developing a feature that touches the proto.

```
2453 |         match msg.get_extra_msg().get_type() {
     |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ pattern `MsgRejectRaftLogCausedByMemoryUsage` not covered
```

PR #914, opened 2022-05-10
proto:add querystat in SyncRegionResponse 

qidi1:
add queryStat in SyncRegionResponse,because we want to add queryStat sync.
Signed-off-by: qidi1 <1083369179@qq.com>


disksing:
Hi, would you like to revert the format changes?

disksing:
closing due to inactivity.

PR #916, opened 2022-05-12
proto: add request_source in Context

you06:
ref pingcap/tidb#34162

To trace the resource usage in TiKV, this PR adds `request_source` in Context. In TiDB, this field will be like "external_select", "external_update", "internal_ddl_add_index", etc. It's just a tag and can also be any string for other clients.


proto/kvrpcpb.proto
```diff
@@ -716,6 +716,9 @@ message Context {
 
     // The informantion to trace a request sent to TiKV.
     tracepb.TraceContext trace_context = 23;
+
+    // The source of the request, will be used as the tag of the trace.
+    string request_source = 24;
```
sticnarf, line 721: If I understand it correctly, it's not used for tracing. Instead, it will be used as the label of some metrics reporting to prometheus.

So I'm hesitating about the field name. I'm a bit more inclined to use a more ambiguous word like "request_label" and it represents its real function better.

you06: The word "label" is a bit common, Prometheus metrics already use "label" for metrics classification, we have "labelType", "labelName", etc. If we name it "request_label", and add "labelLabel" for some metrics, that would be confusing.

sticnarf: I think "source" is okay. We can hint in the comment that this is essentially an extra label in metrics and the user may use it for any actual meaning.

sticnarf (approved):

cfzjywxk (approved):

disksing (approved):

Issue #912, opened 2022-05-10
Remove TiFlash APIs


breezewish:
As a Protobuf of TiKV, it should contain no knowledge about TiFlash.

PR #918, opened 2022-05-17
tiflash: Add TiFlash compaction RPC and messages

breezewish:
Note: The TiFlash Compaction API looks very differently compared to TiKV's. So that we should move these things out when we are going through https://github.com/pingcap/kvproto/issues/912 .


proto/kvrpcpb.proto
```diff
@@ -1151,3 +1151,44 @@ message RawChecksumResponse {
...
+message CompactErrorInvalidStartKey {}
+
+message CompactErrorPhysicalTableNotExist {}
+
+message CompactErrorCompactInProgress {}
+
+message CompactErrorTooManyPendingTasks {}
+
+message CompactRequest {
+    // If specified, the compaction will be performed starting from this start key.
```
JaySon-Huang, line 1173: Add comments about the behavior when the start_key is not set

JaySon-Huang (approved):
LGTM with small comments

proto/kvrpcpb.proto
breezewish: Thanks for the suggestion! More comments have been added to important fields.

PR #920, opened 2022-06-04
gcpb: optimize ListKeySpaces

AmoebaProtozoa:
Signed-off-by: David <8039876+AmoebaProtozoa@users.noreply.github.com>

* removed unnecessary field `new_safe_point` from `UpdateGCSafePointResponse`
* changed `TTL` to `time_to_live`, as all other fields are lower cased, and to avoid confusion with exported vars in go. 
* make `ListKeySpaces` return a stream instead of all at once
* change `space_id` to `uint32` to clarify it's data type


pkg/raft_serverpb/raft_serverpb.pb.go
```diff
@@ -67,8 +67,9 @@ const (
 	ExtraMessageType_MsgCheckStalePeerResponse ExtraMessageType = 3
 	// If leader is going to sleep, it will send requests to all its followers
 	// to make sure they all agree to sleep.
-	ExtraMessageType_MsgHibernateRequest  ExtraMessageType = 4
```
AmoebaProtozoa, line 70: Previous commit didn't `make go`
https://github.com/pingcap/kvproto/commit/6aaebf46630570caab6307cdb2b8f14777383904

PR #922, opened 2022-06-10
Add GetLockWaitHistory rpc

longfangsong:
Signed-off-by: longfangsong <longfangsong@icloud.com>


cfzjywxk:
@longfangsong 
Please resolve the conflicts and if it's reviewable we could change the status of this pr.

ekexium (approved):
Seems go code needs to be generated

cfzjywxk:
@longfangsong 
PTAL and better to add issue reference in the description.

proto/kvrpcpb.proto
```diff
@@ -1119,6 +1119,18 @@ message GetLockWaitInfoResponse {
...
+message GetLockWaitHistoryRequest {
+    Context context = 1;
+    // TODO: There may need some filter options to be used on conditional querying, e.g., finding
+    // the lock waiting status for some specified transaction.
+}
+
+message GetLockWaitHistoryResponse {
+    errorpb.Error region_error = 1;
+    string error = 2;
+    repeated deadlock.WaitForEntry entries = 3;
```
MyonKeminta, line 1131: The `deadlock.WaitForEntry` type doesn't seem to contain a time. Do we need to record the time when the waiting happens?

MyonKeminta (commented):
Rest LGTM

proto/kvrpcpb.proto
sticnarf: Seems not easy? I don't know how to get this time... System time on TiKV is not appropriate.

sticnarf (approved):
LGTM

proto/kvrpcpb.proto
MyonKeminta: Hmmm... I'm still afraid that when a user views the history, the user may want to know approximately when each lock waiting did happen. Without any information about time, it might be a little difficult for a user to figure out which information is helpful for some specific problem. 
Maybe for_update_ts can be a better choice than TiKV's system time? However for_update_ts is not included in the entries either...

cfzjywxk:
@longfangsong 
Please resolve the conflicts.

PR #926, opened 2022-06-21
Add more exec details for txn requests

sticnarf:
I add some more exec details to the responses of the main txn commands.

I want to be a bit conservative at first. It may not be a good idea to transmitting too much information that is not very useful.

Something that is definitely missing is the perf context about writing logs to the raft engine. Because [tikv/raft-engine](https://github.com/tikv/raft-engine) does not provide perf context like RocksDB yet. We have to know what useful metrics there could be before adding them to the protocol.

And another thing I'm considering is to add a threshold in the request. The threshold may be provided by the client, and these detailed information is only returned when the total time exceeds the threshold. I think this should be more flexible than a fixed threshold configured at TiKV but I'm not sure if it is worth...


proto/kvrpcpb.proto
```diff
@@ -922,6 +934,12 @@ message ScanDetailV2 {
 
     // Total number of bytes from block reads.
     uint64 rocksdb_block_read_byte = 7;
+
+    // Total time used for block reads.
+    uint64 rocksdb_block_read_nanos = 9;
```
you06, line 939: Skip `xxx = 8`?

sticnarf: `processed_versions_size = 8` above

you06 (approved):

cfzjywxk:
How should the client decide the threshold? Using the default slow log duration?

cfzjywxk (approved):

sticnarf:
> How should the client decide the threshold? Using the default slow log duration?

A possible way is to maintain the 99% percentile latency of a kind of requests during a day and use that as the threshold...

PR #927, opened 2022-06-22
add tablet support

BusyJay:


proto/raft_serverpb.proto
```diff
@@ -103,6 +103,8 @@ message RegionLocalState {
     PeerState state = 1;
     metapb.Region region = 2;
     MergeState merge_state = 3;
+    // The apply index corresponding to the storage when it's initialized.
```
Connor1996, line 106: apply index？

BusyJay: Yes. tablet index is actually (max applied) index.

Connor1996: how about describing the tablet naming rule here

BusyJay: It's TiKV's implementation details, I would like to make it inside TiKV.

Connor1996 (approved):
LGTM

5kbpers (approved):

PR #929, opened 2022-06-23
added pitr service

YuJuncen:
This is a part of [checkpoint v3](https://github.com/tikv/tikv/issues/12715).
Signed-off-by: Yu Juncen <yujuncen@pingcap.com>


proto/logbackuppb.proto
```diff
@@ -0,0 +1,43 @@
+syntax = "proto3";
```
3pointer, line 1: maybe add the license

proto/logbackuppb.proto
```diff
@@ -0,0 +1,43 @@
...
+
+// The last flush ts with region information.
+message RegionCheckpoint {
+    errorpb.Error err = 1;
+    RegionIdentity region = 2;
+    uint64 checkpoint = 3;
+}
+
+message GetLastFlushTSOfRegionRequest {
+    repeated RegionIdentity regions = 1;
```
3pointer, line 30: shall we add limit here 🤔

3pointer: One `RegionIdentity` has 16 bytes.
if we have 100k region in a store seems 1.5M size in one request is fine.

3pointer (commented):
rest LGTM

proto/logbackuppb.proto
YuJuncen: It seems that other proto files don't have license header too.🤔

3pointer (approved):

PR #951, opened 2022-06-28
[WIP] Keyspace GC Interface

AmoebaProtozoa:


proto/gcpb.proto
```diff
@@ -107,11 +58,9 @@ message UpdateServiceSafePointRequest {
...
-  ResponseHeader header = 1;
-  // update will be successful if ttl < 0 (a removal request)
-  // or if new safe point >= old safe point and new safe point >= gc safe point
-  bool succeeded = 2;
-  uint64 gc_safe_point = 3;
-  uint64 old_safe_point = 4;
-  uint64 new_safe_point = 5;
+  pdpb.ResponseHeader header = 1;
+  // returns the information about the min service safe point after the update
+  bytes service_id = 2;
```
sunxiaoguang, line 63: For what kind of use cases a service id is useful in response?

proto/gcpb.proto
```diff
@@ -107,11 +58,9 @@ message UpdateServiceSafePointRequest {
...
+  int64 TTL = 3;
```
sunxiaoguang, line 64: Likewise, for what reason do we use TTL for?

proto/gcpb.proto
AmoebaProtozoa: It's here to stay consistent with the old UpdateServiceGCSafePoint https://github.com/pingcap/kvproto/blob/08b02befd813c6f5e6629fc24ab249da98e87bb3/proto/pdpb.proto#L709

PR #952, opened 2022-06-30
keyspacepb

AmoebaProtozoa:
ref issue: tikv/pd#5293


proto/keyspacepb.proto
```diff
@@ -0,0 +1,72 @@
...
+  ARCHIVED = 2;
+}
+
+message UpdateKeyspaceConfigRequest {
+  pdpb.RequestHeader header = 1;
+  string name = 2;
+  // specify config name and values for the target keyspace to update.
+  map<string, string> put = 3;
+  // specify which config to delete, deletion will happen after put operations.
+  repeated string delete = 4;
```
ystaticy, line 47: Suggest to separate the update and delete interfaces?
If we combine deletion and update in ```UpdateKeyspaceConfigRequest```, users may not pay attention to the internal execution logic. May be results are inconsistent with expectations or users have few questions about the internal implementation. We cannot control the behavior of users only through code comments.

AmoebaProtozoa: I don't think we have to worry too much about this. 
These gRPC interfaces are mostly for horizontal inter-service communications, whoever interacts with it probably already have some knowledge of internal workings (`keyspace ID`, for example). 
For user-facing interface, there's pd-ctl and RESTful API.
Also if we separate update and delete, we lose the ability to update and delete in a single "transaction", causing behavior to deviate from RESTful API's JSON Merge Patch.

proto/keyspacepb.proto
```diff
@@ -0,0 +1,80 @@
...
+option (gogoproto.sizer_all) = true;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (rustproto.lite_runtime_all) = true;
+
+option java_package = "org.tikv.kvproto";
+
+// Keyspace provides services to manage keyspaces.
+service Keyspace {
+  rpc UpdateKeyspaceConfig (UpdateKeyspaceConfigRequest) returns (UpdateKeyspaceConfigResponse) {}
```
ystaticy, line 18: What scenarios will call ```UpdateKeyspaceConfig```? If the configuration can be updated freely without restrictions and checks , It will make some risks.

proto/keyspacepb.proto
```diff
@@ -0,0 +1,56 @@
...
+
+option (gogoproto.sizer_all) = true;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (rustproto.lite_runtime_all) = true;
+
+option java_package = "org.tikv.kvproto";
+
+// Keyspace provides services to manage keyspaces.
+service Keyspace {
```
disksing, line 17: It seems that service is readonly. I wonder how can we create/update/delete keyspaces?

AmoebaProtozoa: Create/ UpdateState/ UpdateConfig can be done through HTTP API.

nolouch: the client should create a keyspace via HTTP?

AmoebaProtozoa: Target Keyspace should exist prior to client's connection. 
A default keyspace with name `"DEFAULT"` and id `0` is created when KeyspaceManager initialize, if a client (with keyspace enabled) does not provide a keyspace name, if will connect to default keyspace.

nolouch (approved):
lgtm

proto/keyspacepb.proto
```diff
@@ -0,0 +1,56 @@
...
+  pdpb.RequestHeader header = 1;
+  string name = 2;
+}
+
+message LoadKeyspaceResponse {
+  pdpb.ResponseHeader header = 1;
+  KeyspaceMeta keyspace = 2;
+}
+
+message WatchKeyspacesRequest {
```
AndreMouche, line 49: Why we need to watch all key-spaces?

AndreMouche (commented):
Rest LGTM

AndreMouche (approved):
LGTM

PR #955, opened 2022-07-06
build: Generate cpp files under build directory instead of cpp/kvproto

JaySon-Huang:
Generating cpp files under the source directory makes trouble for building under some cases. For example, when we build for cross-platform binary, or building with debug/release mode, the cpp files are generated under the same directory.

Generating cpp files under the build directory is more convenient.

Signed-off-by: JaySon-Huang <tshent@qq.com>


zanmato1984 (approved):
LGTM

PR #963, opened 2022-07-23
proto: add lock_if_fields in PessimisticLockRequest

TonsnakeLin:
add lock_if_fields in PessimisticLockRequest  and tikv locks the record when it exists.


cfzjywxk:
@TonsnakeLin 
For demo test the dev branch could be used, after that it could be decided if the change should be merged into master branch.

TonsnakeLin:
> @TonsnakeLin For demo test the dev branch could be used, after that it could be decided if the change should be merged into master branch.

Yes, it's better to demo test on the dev branch, but it  brings so many extra jobs. it's not so effective. Hmmm, we should make develop more happier.

disksing:
do we want to merge this PR now? I think we can test based on the branch before merge it to master.

sticnarf:
What about just using the `isolation_level` in the `Context` to decide the behavior of pessimistic locking?

TonsnakeLin:
> What about just using the `isolation_level` in the `Context` to decide the behavior of pessimistic locking?
the `isolation_level`  is generally used to controlling the read behavior, but we want to control the lock behavior at RC isolation . 
Use the `isolation_level` in the `Context` makes code simple, and this feature is indeed related to isolation. Do we think it is a better choice? @cfzjywxk @ekexium  If so, I will fix it.

sticnarf:
Now, I think a new field is clearer. And it may be necessary if there is a case when we also need different behaviors under RC.

By the way, I prefer to call it `lock_only_if_exists`.

cfzjywxk:
Adding a new field is clearer I think. If there are no other uncertainties about this feature we could start to change the related pr status from draft to reviewable and start to merge now. @TonsnakeLin

sticnarf (approved):

cfzjywxk (approved):

PR #968, opened 2022-08-10
raft_serverpb: trace peers' availability info

ethercflow:


proto/raft_serverpb.proto
```diff
@@ -117,10 +117,18 @@ enum ExtraMessageType {
...
+    bool miss_data = 1;
+    uint64 conf_ver = 2;
 }
 
 message ExtraMessage {
     ExtraMessageType type = 1;
     uint64 premerge_commit = 2;
-    repeated metapb.Peer check_peers = 3;
+    ExtraState extra_state = 3;
+    repeated metapb.Peer check_peers = 4;
```
disksing, line 133: no need to consider backward capatibility?

ethercflow: Should consider, thanks!

proto/raft_serverpb.proto
```diff
@@ -117,10 +117,18 @@ enum ExtraMessageType {
     MsgHibernateRequest = 4;
     MsgHibernateResponse = 5;
     MsgRejectRaftLogCausedByMemoryUsage = 6;
+    MsgCollectPeerApplyIndexProgressRequest = 7;
```
BusyJay, line 120: Wrong name.

proto/raft_serverpb.proto
```diff
@@ -117,10 +117,18 @@ enum ExtraMessageType {
...
+    MsgCollectPeerApplyIndexProgressResponse = 8;
+}
+
+message ExtraState {
```
BusyJay, line 124: Bad name.

proto/raft_serverpb.proto
```diff
@@ -117,10 +117,17 @@ enum ExtraMessageType {
     MsgHibernateRequest = 4;
     MsgHibernateResponse = 5;
     MsgRejectRaftLogCausedByMemoryUsage = 6;
+    MsgCollectPeerApplyIndexProgressRequest = 7;
+    MsgCollectPeerApplyIndexProgressResponse = 8;
```
Connor1996, line 121: why need a response

proto/raft_serverpb.proto
ethercflow: How about PeerExtraState?

proto/raft_serverpb.proto
ethercflow: Oh, What I thought was that when the leader sends `MsgAppend` or `MsgHeartbeat`, it sends a `MsgCollectPeerApplyIndexProgressRequest`, and then other peers receive this information and return `MsgCollectPeerApplyIndexProgressResponse`.

But it should be that when non-leader peers send `MsgAppendResp` or `MsgHeartBeatResp`, they carry the `extra state` information, and when the leader receives `MsgAppendResp` or `MsgHeartBeatResp`, it processes the received information. So no need for an `extra resp type` right?

proto/raft_serverpb.proto
ethercflow: Fixed, thanks!

proto/raft_serverpb.proto
ethercflow: Removed, thanks!

Connor1996 (approved):
LGTM

proto/raft_serverpb.proto
```diff
@@ -117,10 +117,13 @@ enum ExtraMessageType {
     MsgHibernateRequest = 4;
     MsgHibernateResponse = 5;
     MsgRejectRaftLogCausedByMemoryUsage = 6;
+    MsgTracePeerAvailabilityInfo = 7;
```
BusyJay, line 120: Can it be shorter?

ethercflow: I don't think it can, I looked at synonyms and there is no short :(

BusyJay: Does it need a request type and a response type?

ethercflow: No, this message will be sent by non-leader peers when replying `MsgAppendResp` or `MsgHeartBeatResp`, leader use this type to collect infos.

BusyJay: So it will be attached to every messages? I don't think it is necessary.

ethercflow: Yes, as every `MsgAppendResp` as applied index changes, or `MsgHeartBeatResp` if there's long time no write.

BusyJay: I don't object to report the response periodically, but I don't think it's necessary to be attached to heartbeat or append response. It should be queried on demand.

Connor1996: If it's queried on demand, it's still triggered on leader periodically. Seems unnecessary

BusyJay: Once it's recovered from not ready to ready and vice versa, then no need to query anymore until there the role is changed again.

proto/raft_serverpb.proto
```diff
@@ -125,10 +125,13 @@ enum ExtraMessageType {
...
 }
 
 message ExtraMessage {
     ExtraMessageType type = 1;
     uint64 premerge_commit = 2;
     repeated metapb.Peer check_peers = 3;
+    bool miss_data = 4;
+    uint64 safe_ts = 5;
```
BusyJay, line 136: If it's not used, it should be removed.

ethercflow: Fixed, thanks!

ethercflow:
@BusyJay Msg has been split into request and response, also added a new PeerState `PeerState::Unavailable` as discussed earlier in the group.

proto/raft_serverpb.proto
```diff
@@ -106,6 +106,7 @@ enum PeerState {
     Applying = 1;
     Tombstone = 2;
     Merging = 3;
+    Unavailable = 4;
```
Connor1996, line 109: add a comment for this

Connor1996 (commented):
rest LGTM

proto/raft_serverpb.proto
```diff
@@ -139,10 +140,13 @@ enum ExtraMessageType {
     MsgHibernateRequest = 4;
     MsgHibernateResponse = 5;
     MsgRejectRaftLogCausedByMemoryUsage = 6;
+    MsgTracePeerAvailabilityInfoRequest = 7;
```
BusyJay, line 143:
```suggestion
    MsgAvailabilityRequest = 7;
```

ethercflow: Renamed, thanks!

BusyJay (approved):

PR #970, opened 2022-08-18
ebs br: remove recover alloc id GRPC api

D3Hunter:
previously we're using rpc to forward to leader, but http request can do it too, so there's no need to have an rpc api


fengou1 (approved):
gRPC alloc id can be removed since the BR will read all region meta.
during the second phase of recovery, we collect all ID (peer id and region id), calc max ID  from region meta, and reset pd alloc id.

3pointer (approved):

PR #972, opened 2022-08-22
fix a relative path issue

Kelo007:
The function `collect` assumes that the current directory is `./proto`, so it uses `mkdir -p ../pkg/$base_name` to make a directory.
It's true in lines 40-44, because it has already cd to proto.
However, it's obviously not true in lines 38-39.

The issue can occur when remove the directory `./pkg/eraftpb`.


disksing (approved):

disksing:
Thanks for your contribution @Kelo007

PR #975, opened 2022-08-25
Add for witness in snapshot meta

Connor1996:
Generate an empty snpashot for wintess is okay, but better add a new field in snapshot meta to make sure other peers won't apply it by mistake.


BusyJay:
How can old version peer detect such error?

Connor1996:
> How can old version peer detect such error?

Witness is enabled only when the cluster version meets the feature needs, so old version peer won't encounter snapshot generated for witness

disksing:
PTAL @BusyJay

proto/raft_serverpb.proto
```diff
@@ -45,6 +45,8 @@ message SnapshotMeta {
     repeated SnapshotCFFile cf_files = 1;
     // true means this snapshot is triggered for load balance
     bool for_balance = 2;
+    // true means this is an empty snapshot for witness
```
BusyJay, line 48: Is this still necessary after using file less snapshot?

Connor1996: yes, need it to decide how to handle the snapshot

BusyJay (approved):

PR #979, opened 2022-08-31
add wait apply for region sync to last index

fengou1:
What changed:
add a new interface wait apply to wait raft log apply reach the target index

Why not reuse existed recoverCmd interface?
to ensure the raft log apply to last index. we have do as follower
1. assign leader and wait leader apply to last index
2. wait follower wait apply to last index

step 1 is precondition for step 2

also there are other solution, likes: leader to check follower applied index, it impact raftstore, new message may need.
Signed-off-by: fengou1 <feng.ou@pingcap.com>


3pointer (approved):

proto/recoverdatapb.proto
```diff
@@ -30,18 +30,28 @@ message RegionMeta {
...
 }
 
+// wait apply to last index
+ message WaitApplyRequest {
+     uint64 store_id = 1;
+ }
+
+ message WaitApplyResponse {
+     Error error = 1;
+     uint64 store_id = 2;
```
D3Hunter, line 52: what's difference between this store_id and request store id?

D3Hunter (commented):
rest lgtm

D3Hunter (approved):

PR #984, opened 2022-09-06
raft_serverpb: add StoreRecoverState

5kbpers:
Signed-off-by: 5kbpers <hustmh@gmail.com>


proto/raft_serverpb.proto
```diff
@@ -75,6 +75,10 @@ message StoreIdent {
     kvrpcpb.APIVersion api_version = 3;
 }
 
+message StoreRecoverState {
+    uint64 seqno = 1;
```
BusyJay, line 79: Can you add some comments to explain the struct and its field?

BusyJay (approved):

PR #985, opened 2022-09-08
check admin: check store has pending admin reqeusts

3pointer:


hicqu (approved):

proto/brpb.proto
```diff
@@ -347,8 +348,21 @@ message BackupResponse {
...
+    metapb.Region region = 2;
+    bool has_pending_admin = 3;
+}
+
 service Backup {
     rpc backup(BackupRequest) returns (stream BackupResponse) {}
+    // CheckAdmin used for snapshot backup. before we start snapshot for a TiKV. 
+    // we need stop all schedule first and make sure all in-flight schedule has finished.  
+    // this rpc check all pending conf change for leader.
+    rpc CheckAdmin(CheckAdminRequest) returns (stream CheckAdminResponse) {}
```
D3Hunter, line 365: name it to HasRegionSchedule?

3pointer: I think schedule is also not clear, we need check split/merge rather than schedule 🤔
CheckPendingAdmin maybe better?

D3Hunter: or `CheckPendingAdminOp`, don't know what it mean by `pending admin`

3pointer: CheckPendingAdminOp is ok for me, WDYT? cc @hicqu

3pointer: change it to CheckPendingAdminOp for now, if it's not ok, we can change it in future.

D3Hunter (approved):

PR #990, opened 2022-09-29
raft_cmdpb, raft_serverpb: make flashback state persist in

HuSharp:

Ref https://github.com/tikv/tikv/issues/13519.


JmPotato (approved):

proto/raft_serverpb.proto
```diff
@@ -120,6 +122,7 @@ message RegionLocalState {
     MergeState merge_state = 3;
     // The apply index corresponding to the storage when it's initialized.
     uint64 tablet_index = 4;
+    bool is_in_flashback  = 5;
```
JmPotato, line 125: We can move this filed into `metapb.Region`, so that there is no need to add two different fields in `RegionLocalState` and `SnapshotMeta`.

HuSharp: Okay, I'll make the change.

disksing (approved):

PR #997, opened 2022-10-31
Add info about last change to LockInfo

sticnarf:
Refer to https://github.com/tikv/tikv/issues/13694

`last_change_ts` and `versions_to_last_change` will be recorded in the pessimistic locks and prewritten locks. These two fields are not useful for TiDB but the in-memory pessimistic locks are serialized in the format of protobuf. So, we need these two fields in kvproto.

The fields are also added into `MvccInfo` to make debugging easier.


MyonKeminta (approved):

youjiali1995:
> in-memory pessimistic locks are serialized in the format of protobuf

Where? Seems they are serialized to `Lock`. https://github.com/tikv/tikv/blob/d9fe2ffd78f4db5ec2141c98e019d14ba980f121/components/raftstore/src/store/fsm/peer.rs#L3312

sticnarf:
> > in-memory pessimistic locks are serialized in the format of protobuf
> 
> Where? Seems they are serialized to `Lock`. https://github.com/tikv/tikv/blob/d9fe2ffd78f4db5ec2141c98e019d14ba980f121/components/raftstore/src/store/fsm/peer.rs#L3312

Oops. I happened to think about it wrong. Somehow I thought `lock.to_bytes()` serialized it into `LockInfo` protobuf...

PR #999, opened 2022-11-01
add lagging_peers field

ethercflow:
We need promote the witness to voter when region has lagging behind peer. The current pending peers include new peers scheduled by pd, which are also considered pending. Therefore, this item needs to be added to distinguish.

On the TiKV side, the judgment will be made according to the value of process.matched. If it is non-zero and < truncated_idx, it will be regarded as lagging peers. The logic of pending peers will not be modified.


BusyJay:
> We need promote the witness to voter when region has lagging behind peer.

What if the peer is slowed down by applying entries? Is it possible that pd will keep changing peers from witness to non-witness occasionally?

And if a follower is isolated before new leader is elected, its progress is always 0. What should PD do in this case?

ethercflow:
> What if the peer is slowed down by applying entries? Is it possible that pd will keep changing peers from witness to non-witness occasionally?

I think it's possible, currently I'm not sure what the jitter of pending peers is due to slow apply, and whether a higher threshold is needed for lagging. I plan to observe it when doing system testing.  cc @Connor1996

ethercflow:
> And if a follower is isolated before new leader is elected, its progress is always 0. What should PD do in this case?

It's considered as down peer.

ethercflow:
> What if the peer is slowed down by applying entries? Is it possible that pd will keep changing peers from witness to non-witness occasionally?

@BusyJay  I thought twice and realized I didn't understand your concern. What you mean is that if a region has multiple witnesses, when a lagging peer is found, and after converting one witness to non-witness, since the lagging peer has not recovered, will pd continue to convert the remaining witnesses to non-witness?

BusyJay:
> @BusyJay I thought twice and realized I didn't understand your concern. What you mean is that if a region has multiple witnesses, when a lagging peer is found, and after converting one witness to non-witness, since the lagging peer has not recovered, will pd continue to convert the remaining witnesses to non-witness?

For example, a is leader, b is follower, c is witness. And b is slow because it needs to apply logs. So PD decides to make b as witness and c as non-witness. After that, c is slow because now it needs to apply logs, so PD make c witness again. It is trapped in a loop and no way to get out.

ethercflow:
> For example, a is leader, b is follower, c is witness. And b is slow because it needs to apply logs. So PD decides to make b as witness and c as non-witness. After that, c is slow because now it needs to apply logs, so PD make c witness again. It is trapped in a loop and no way to get out.

When we find b is a lagging peer, PD will create an op to promote c to non-witness and return. The op is only confirmed to be completed after the non-witness of the apply snapshot is completed, so there will be no ops that turn b into a witness until then. Now c is already a normal voter, won't snapshot be needed again in the short term?

BusyJay:
> so there will be no ops that turn b into a witness until then.

Timing doesn't matter. The key point here is witness and non-witness will be kept scheduled around. The worst case is that if non-witness becomes pending, there is only one healthy replica in the group and one node lost can cause permanent data loss.

ethercflow:
> > so there will be no ops that turn b into a witness until then.
> 
> Timing doesn't matter. The key point here is witness and non-witness will be kept scheduled around. The worst case is that if non-witness becomes pending, there is only one healthy replica in the group and one node lost can cause permanent data loss.

So, it looks meaningless to promote witness to non-witness under these two cases, right?
- Promotes the witness to voter when region has one down peer
- Promotes the witness to voter when region has lagging behind peer

cc @Connor1996

Connor1996:
> > @BusyJay I thought twice and realized I didn't understand your concern. What you mean is that if a region has multiple witnesses, when a lagging peer is found, and after converting one witness to non-witness, since the lagging peer has not recovered, will pd continue to convert the remaining witnesses to non-witness?
> 
> For example, a is leader, b is follower, c is witness. And b is slow because it needs to apply logs. So PD decides to make b as witness and c as non-witness. After that, c is slow because now it needs to apply logs, so PD make c witness again. It is trapped in a loop and no way to get out.

PD won't make b witness as there is a lagging behind peer in region.

ethercflow:
> > > @BusyJay I thought twice and realized I didn't understand your concern. What you mean is that if a region has multiple witnesses, when a lagging peer is found, and after converting one witness to non-witness, since the lagging peer has not recovered, will pd continue to convert the remaining witnesses to non-witness?
> > 
> > 
> > For example, a is leader, b is follower, c is witness. And b is slow because it needs to apply logs. So PD decides to make b as witness and c as non-witness. After that, c is slow because now it needs to apply logs, so PD make c witness again. It is trapped in a loop and no way to get out.
> 
> PD won't make b witness as there is a lagging behind peer in region.

But it still can't handle: 
> The worst case is that if non-witness becomes pending, there is only one healthy replica in the group and one node lost can cause permanent data loss.

right?

eg: a is leader, b is a lagging follower, c is a pending non-witness, there is still only one healthy replica in the group?

Connor1996:
I don't get the point. `a is leader, b is a lagging/pending/down follower, c is a pending non-witness` is an expected state, then the pending non-witness would become available finally. It's unavoided with the existence of witness. That's why we need to promote the witness to nonwitness to make sure there is not only one healthy replica in the group.

ethercflow:
> then the pending non-witness would become available finally.

My doubt is before it becomes available, if the leader node lost can still cause permanent data loss?

BusyJay:
> PD won't make b witness as there is a lagging behind peer in region.

Timing doesn't matter at all. The point is lagging behind is possible because of different nodes performs differently, especially technically witness should be faster than non-witness. There are two different problems:

1. loop scheduling, 3 healthy non-witness -> 1 witness + 2 healthy non-witness -> 1 witness + 1 lagging behind non-witness + 1 healthy witness -> 2 healthy non-witness + 1 lagging behind non-witness -> 3 healthy non-witness -> ...
2. Allow lagging behind majority means sacrifices durability.

Connor1996:
> > PD won't make b witness as there is a lagging behind peer in region.
> 
> Timing doesn't matter at all. The point is lagging behind is possible because of different nodes performs differently, especially technically witness should be faster than non-witness. There are two different problems:
> 
> 1. loop scheduling, 3 healthy non-witness -> 1 witness + 2 healthy non-witness -> 1 witness + 1 lagging behind non-witness + 1 healthy witness -> 2 healthy non-witness + 1 lagging behind non-witness -> 3 healthy non-witness -> ...
> 2. Allow lagging behind majority means sacrifices durability.

Let me change a little of the definition:
- pending peer: a new adding peer, a new non-witness, a peer whose matched_idx < truncated_idx (original definition)
- lagging peer: a peer whose commit_idx - matched_idx(except 0) > threshold 
1. it's expected, we have no choose, but the lagging won't happen so easily. 
2. It doesn't affect durability. Note that when there is a witness, raft log gc won't be advanced until the log is replicated to all peers. So even with a lagging peer, and the leader is down, the witness has enough log to help the lagging peer catch up. But raft log may accumulate a lot and waste disk space. That's why we need to promote witness to non-witness, then raft log gc can be advanced as usual.

BusyJay:
>     1. it's expected, we have no choose, but the lagging won't happen so easily.

No, it's not expected. This can result in instability as data are kept removed and ingested, it will lead to useless high compaction in the end. And this will be more likely with the new definition.

>     But raft log may accumulate a lot and waste disk space. That's why we need to promote witness to non-witness, then raft log gc can be advanced as usual.

Then the policy is not about the lagging behind, instead should be whether too many logs are still reserved because of witness mechanism, which should be observable by looking into raft engine's metrics. And the region should be put into blocklist to avoid scheduling loop.

Connor1996:
Scheduling loop can't happen so easily, when 1 leader + 1 lagging behind follower + 1 witness -> 1 leader + 1 lagging behind follower + 1 non-witness, the lagging behind follower can't catch up logs quickly, it must has something wrong with it. Then the lagging behind follower would avoids the non-witness->witness until it's marked down and removed somehow.

Connor1996:
> Then the policy is not about the lagging behind, instead should be whether too many logs are still reserved because of witness mechanism, which should be observable by looking into raft engine's metrics.

Then you have no clue when you can add back the witness

BusyJay:
> it must has something wrong with it.

Not necessary. As explained above, witness is technically faster than non-witness. Once witness is promoted to non-witness, it also becomes slow, so the raft group is slow down and allow lagging behind non-witness to catch up logs.

> when you can add back the witness

There should be ttl in blocklist. Once the ttl is met and the pending raft logs remains in a low ratio, non-witness can be set to witness again.

Connor1996:
Then we can just keep using lagging peer, and non-witness can't be set to witness again until ttl is met and there is no lagging/pending/down peer

BusyJay:
I think the real problems here is to avoid too many accumulated logs. Lagging behind is a relative state that somehow reflects the size of logs but not accurate. The definition `commit_idx - matched_idx(except 0) > threshold` seems quite arbitrary and hard to reason what `threshold` should be.

Have you tested the definition and does it well serve the purpose?

Connor1996:
The threshold is `raft_log_gc_count_limit` and `raft_log_gc_size_limit`. It's an existing definition representing too many accumulated logs

PR #1001, opened 2022-11-02
Update comment of `kvrpcpb.APIVersion` according to latest implementation

pingyu:
Update comment of `kvrpcpb.APIVersion` according to latest implementation.


pingyu:
@iosmanthus @haojinming @zeminzhou PTAL, thanks~

proto/kvrpcpb.proto
```diff
@@ -791,48 +791,54 @@ message Context {
 // The API version the server and the client is using.
 // See more details in https://github.com/tikv/rfcs/blob/master/text/0069-api-v2.md.
 enum APIVersion {
-    // Mainly for TxnKV and not safe to use RawKV along with TxnKV.
-    //
-    // V1 server only accepts V1 requests. Except that the V1 raw requests with TTL
-    // will be rejected.
+    // `V1` is mainly for TiDB & TxnKV, and is not safe to use RawKV along with the others.
```
haojinming, line 794: Is V1 safe to use `TxnKV` along with `TiDB`? In case user may contruct the txn key with same coding method with TiDB. I think user should not use mixed types in V1 mode.

zeminzhou (approved):
LGTM~

proto/kvrpcpb.proto
disksing: It does carry some risk, but we do support it.

disksing (approved):

proto/kvrpcpb.proto
```diff
@@ -791,48 +791,54 @@ message Context {
...
-    // --------------------------------------
-    // | 0x12 0x34 0x56 | 0x00 (0b00000000) |
-    // --------------------------------------
+    // `V2` use new encoding for RawKV & TxnKV to support more features.
     //
-    // As shown in the example below, the least significant bit of the meta flag
-    // indicates whether the value contains 8 bytes expire ts at the very left to the
-    // meta flags.
+    // Key Encoding:
+    //  TiDB: start with `m` or `t`, the same as `V1`.
```
ystaticy, line 812: Because `m` and `t` are added in the logic of TiDB,
We can say that : TiDB uses the v1 client, and user key starts with `m` and `t` ?

PR #1004, opened 2022-11-03
add a new admin cmd

ethercflow:
Add a new admin command discussed in https://github.com/tikv/tikv/pull/12972#discussion_r1000434376


proto/pdpb.proto
```diff
@@ -433,6 +433,10 @@ message SplitRegion {
     repeated bytes keys = 2;
 }
 
+message SwitchWitness {
+    metapb.Peer peer = 1;
```
BusyJay, line 437: Seems overwhelming to use `Peer`. How about using ID and flag?

ethercflow: Sure, using ID and flag instead, thanks!

proto/pdpb.proto
```diff
@@ -433,6 +433,11 @@ message SplitRegion {
     repeated bytes keys = 2;
 }
 
+message SwitchWitness {
+    uint64 id = 1;
```
Connor1996, line 437:
```suggestion
    uint64 peer_id = 1;
```

ethercflow: Fixed, thanks!

Connor1996 (approved):
LGTM

proto/raft_cmdpb.proto
```diff
@@ -220,7 +220,8 @@ message RollbackMergeRequest {
 message RollbackMergeResponse {}
 
 message SwitchWitnessRequest {
-    metapb.Peer peer = 1;
+    uint64 id = 1;
```
BusyJay: Should it be repeated?

proto/raft_cmdpb.proto
```diff
@@ -220,7 +220,8 @@ message RollbackMergeRequest {
...
+    bool isWitness = 2;
```
BusyJay: Inconsistent style

proto/raft_cmdpb.proto
ethercflow: I think should be. eg: distributed across different stores

proto/raft_cmdpb.proto
ethercflow: consistent, maybe your browser didn't refresh :)

proto/raft_cmdpb.proto
ethercflow: Fixed, thanks!

proto/pdpb.proto
```diff
@@ -433,6 +433,11 @@ message SplitRegion {
     repeated bytes keys = 2;
 }
 
+message SwitchWitness {
+    repeated uint64 peer_id = 1;
```
BusyJay, line 437: Seems extract them as a struct and make the struct repeated is a better style. But this is also OK.

BusyJay (approved):

proto/pdpb.proto
ethercflow: Ok, I plan to change it to `SwitchWitness` and `BatchSwitchWitness` like `Split` and `BatchSplit`.

proto/raft_cmdpb.proto
```diff
@@ -219,6 +219,19 @@ message RollbackMergeRequest {
 
 message RollbackMergeResponse {}
 
+message SwitchWitnessRequest {
+    uint64 peer_id = 1;
+    bool is_witness = 2;
+}
+
+message SwitchWitnessResponse {}
```
BusyJay, line 227: Seems not use at all.

ethercflow: Removed, thanks!

PR #1009, opened 2022-11-11
log-backup: added subscribe flush rpcs

YuJuncen:


proto/logbackuppb.proto
```diff
@@ -34,10 +34,25 @@ message GetLastFlushTSOfRegionResponse {
     repeated RegionCheckpoint checkpoints = 1;
 }
 
+message SubscribeFlushEventRequest {
+    string client_id = 1;
```
joccau, line 38: Is this `client_id` is "store_id" ?

YuJuncen: This can be a random string, just for debugging usage.

Leavrth (approved):

PR #1017, opened 2022-11-23
update grpc to 0.10.4

tabokie:
Signed-off-by: tabokie <xy.tao@outlook.com>

For https://github.com/tikv/grpc-rs/pull/597


disksing (approved):

BusyJay:
Hmm, actually you don't need to update this. cargo will automatically resolve 0.10.0 to 0.10.4.

PR #1025, opened 2022-12-05
Introduce i64 priority for better compatibility

ethercflow:
ref tikv/tikv#12876


Connor1996 (approved):
LGTM. Let's merge it when raft-rs is updated

include/eraftpb.proto
```diff
@@ -80,6 +80,12 @@ message Message {
     uint64 reject_hint = 11;
     bytes context = 12;
     uint64 priority = 14;
+    // If this new field is not set, then use the above old field; otherwise
+    // use the new field. When broadcasting request vote, both fields are
+    // set if the priority is larger than 0. This change is not a fully
+    // compatible change, but it makes minimal impact that only new priority
+    // is not recognized by the old nodes during rolling update.
+    int64 priority_v2 = 15;
```
BusyJay, line 88: I suggest to change this to `priority` and the old one to `deprecated_priority`.

BusyJay (approved):

PR #1026, opened 2022-12-08
MPP: add some field to support TiFlash stale read

hehechen:
https://pingcap.feishu.cn/docx/HONRdCwfloUaxbxXRWucaHognmb


proto/mpp.proto
```diff
@@ -17,6 +17,9 @@ message TaskMeta {
 	int64 task_id = 2; // if task id is -1 , it indicates a tidb task.
 	int64 partition_id = 3; // Only used for hash partition
 	string address = 4; // target address of this task.
+	int64 query_ts = 5; // timestamp when query starting to execute, used for TiFlash miniTSO schedule.
```
windtalker, line 20: Is there a guarantee that `query_ts` generated in the same TiDB node always different?

hehechen: > Is there a guarantee that `query_ts` generated in the same TiDB node always different?

No, it's just the system clock timestamp so that MPP schedule in TiFlash can use it to schedule. We need to use query_ts + local_query_id + server_id to generate a unique query_id.

PR #1035, opened 2023-01-10
support watch resource group

HuSharp:


disksing (approved):

ekexium:
I think changes that break compatibility had better be merged together with dependent PRs. Otherwise it may block other changes

HuSharp:
> I think changes that break compatibility had better be merged together with dependent PRs. Otherwise it may block other changes

Sry...My fault, should have made a draft label...
I'll do a tikv fix proto firstly

proto/pdpb.proto
```diff
@@ -103,33 +103,43 @@ service PD {
...
-    repeated string names = 1;
+    string config_path = 1;
 }
 
 message LoadGlobalConfigResponse {
     repeated GlobalConfigItem items = 1;
+    int64 revision = 2;
+}
+
+enum ItemKind {
```
rleungx, line 133: How about calling it EventType or something else which can represent that it is an action?

HuSharp: You are right! from that way we can sync with Etcd

proto/pdpb.proto
```diff
@@ -103,33 +103,43 @@ service PD {
...
 
 message WatchGlobalConfigResponse {
     repeated GlobalConfigItem changes = 1;
+    int64 revision = 2;
 }
 
 message StoreGlobalConfigRequest {
-    repeated GlobalConfigItem changes = 1;
+    string config_path = 1;
+    repeated GlobalConfigItem changes = 2;
```
Connor1996, line 117: breaking change, shouldn't change the field number

HuSharp: This interface was previously used by TopSQL as a configuration center control switch, but then chose another way.
tikv/pd/tidb are not use it.

PR #1038, opened 2023-01-11
Support tso service

rleungx:
Signed-off-by: Ryan Leung <rleungx@gmail.com>


rleungx:
@binshi-bing Would you like to take a look?

proto/tsopb.proto
```diff
@@ -0,0 +1,42 @@
+syntax = "proto3";
+package tsopb;
```
binshi-bing, line 2: According to our last sync, my understanding is that we'll provide these APIs as storage interface used by tso service internal.

proto/tsopb.proto
```diff
@@ -0,0 +1,42 @@
...
+import "gogoproto/gogo.proto";
+import "rustproto.proto";
+import "pdpb.proto";
+
+option (gogoproto.sizer_all) = true;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (rustproto.lite_runtime_all) = true;
+
+service Tso {
```
binshi-bing, line 13: This service is to provide TSO storage. Shall we call it TsoStorage to differentiate it from the TSO service which provide GetTSs?

binshi-bing (commented):
Though we don't need this change for now, this is my first review for PD.

PR #1039, opened 2023-01-12
resoucemanager: add more comments

nolouch:
Signed-off-by: nolouch <nolouch@gmail.com>


proto/resource_manager.proto
```diff
@@ -148,7 +150,9 @@ enum GroupMode {
 message ResourceGroup {
   string name = 1;
   GroupMode mode = 2;
+  // Used in RU mode, group settings with WRU/RRU etc resource abstract unit.
   GroupRequestUnitSettings r_u_settings = 3;
```
Connor1996, line 154: how about ru_mode_setting and raw_mode_setting

nolouch: How about now? `mode` already a field so I not used it.

Connor1996 (approved):
LGTM

PR #1041, opened 2023-01-17
pdpb: update global config definition

HuSharp:

- modify item kind to represent that it is an action
- remove item error


JmPotato (approved):

proto/pdpb.proto
```diff
@@ -163,10 +160,9 @@ enum ErrorType {
     ALREADY_BOOTSTRAPPED = 4;
     INCOMPATIBLE_VERSION = 5;
     REGION_NOT_FOUND = 6;
-    GLOBAL_CONFIG_NOT_FOUND = 7;
-    DUPLICATED_ENTRY = 8;
-    ENTRY_NOT_FOUND = 9;
-    INVALID_VALUE = 10;
+    DUPLICATED_ENTRY = 7;
```
glorv, line 163:
```suggestion
    reserved 7;
```
We should just deprecate this variant but keep other variants unchanged so we can be compatible with old versions.

glorv (commented):
rest LGTM

proto/pdpb.proto
HuSharp: get it, Done.

proto/pdpb.proto
```diff
@@ -163,10 +161,12 @@ enum ErrorType {
     ALREADY_BOOTSTRAPPED = 4;
     INCOMPATIBLE_VERSION = 5;
     REGION_NOT_FOUND = 6;
-    GLOBAL_CONFIG_NOT_FOUND = 7;
+    reserved 7; reserved "GLOBAL_CONFIG_NOT_FOUND";
     DUPLICATED_ENTRY = 8;
     ENTRY_NOT_FOUND = 9;
     INVALID_VALUE = 10;
+    // watch etcd revision supplied does not match the current etcd revision
+    REVISION_MISMATCH = 11;
```
glorv, line 169:
```suggestion
    // required watch revision is smaller than current compact/min revision.
    DataCompacted = 11;
```

glorv (approved):
LGTM

nolouch (approved):

PR #1047, opened 2023-01-28
add encryption info to snapshot chunk

tabokie:


BusyJay:
So why is encryption done using absolute path? If it's using relative path, then copy directory can be sufficient for encrypted db. It should reduce a lot of complexity and resource usage.

tabokie:
You lost me there. The target machine doesn't have the keys stored on source machine. I don't see how relative path can avoid passing keys around.

BusyJay:
How can different machine use different keys? And why would they do so? In practice, they should use the same master key and generate different data key to do the actual encryption. What we should do is sending the snapshot with data key to different machines.

tabokie:
> In practice, they should use the same master key and generate different data key to do the actual encryption.

(1) master key (and data keys) can be rotated, and the rotation will certainly not be in sync between multiple instances (2) the data key in use is chosen from a list of provided data keys, its rotation is determined by local clock which will not be in sync between instances too.

> What we should do is sending the snapshot with data key to different machines.

That's the purpose of this PR, to send data with keys.

BusyJay:
> (1) master key (and data keys) can be rotated, and the rotation will certainly not be in sync between multiple instances 
> (2) the data key in use is chosen from a list of provided data keys, its rotation is determined by local clock which will not be in sync between instances too.

I don't see why need to sync them. The purpose is sending data key with encrypted files. And then using the *same* master key to decrypt the data key and the file.

> > What we should do is sending the snapshot with data key to different machines.
> 
> That's the purpose of this PR, to send data with keys.

Can you explain more in the PR description or code comment? It's hard to know what the PR is going to be implemented. For example, what's in the data field? Are they going to be encrypted or not? When to fill `encryption_info` and what's the mapping?

tabokie:
I still don't understand the relevance to master key and relative directory. The proposed flow is to send encrypted data and its encrypted key, at target machine, it inserts the key into its local key directory, and use the encrypted snapshot as is. The only downside is the network compression (if any) will be crippled.

BusyJay:
There is nothing between master key and relative directory. TiKV currently tracks encrypted data files using the absolute path of the files, so **data key** is mapping to file using absolute path. It causes two problems:
- sending a snapshot is not a simple cp via network, instead it either needs to decrypt the data before being sent or using the method in this PR;
- applying a snapshot can't be a simple mv.

If using relative directory, both cp via network and mv doesn't change the path between data files and data keys, so nothing needs to be changed.

tabokie:
> sending a snapshot is not a simple cp via network, instead it either needs to decrypt the data before being sent or using the method in this PR;

I don't understand how there's any alternative besides this two. Target instance knows nothing about the keys of files from source instance, as explained in https://github.com/pingcap/kvproto/pull/1047#issuecomment-1407320558. There's nothing "relative path" can do to magically pass information between two machines.

> applying a snapshot can't be a simple mv.

This is an entirely different problem. But for the sake of discussion, I don't see how relative path can avoid key transferring too. Data keys are not stored with the data, it's stored centrally in one file. If the absolute path is changed, the relative path is changed too (I'm assuming you are referring to the path of data files relative to encryption key files).

However, if you are referring to the relative path within data dir, then it's possible to make encryption keys shared between multiple subdirectories. For example, data/db/1_2 and data/snapshot/1_2. It would require changing the naming schema a bit.

BusyJay:
Now it's using snapshot cache, encryption info should be added to other message.

BusyJay (approved):

PR #1049, opened 2023-01-29
resource group: add pu mode

disksing:


proto/resource_manager.proto
```diff
@@ -93,6 +98,8 @@ message TokenBucketResponse {
   repeated GrantedRUTokenBucket granted_r_u_tokens = 2;
   // Raw mode
   repeated GrantedRawResourceTokenBucket granted_resource_tokens = 3;
+  // PU mode
+  double granted_p_u_tokens = 4;
```
CabinfeverB, line 102: Should use `GrantedPUTokenBucket`?

disksing: yes!

PR #1050, opened 2023-01-31
[Resource Manager] Unify RRU and WRU into RU

CabinfeverB:
Signed-off-by: Cabinfever_B <cabinfeveroier@gmail.com>


proto/resource_manager.proto
```diff
@@ -95,21 +95,20 @@ message TokenBucketResponse {
...
 
-message GrantedRUTokenBucket {
-  RequestUnitType type = 1;
+message GrantedRawResourceTokenBucket {
+  RawResourceType type = 1;
   TokenBucket granted_tokens = 2;
   int64 trickle_time_ms = 3;
 }
 
 enum RequestUnitType {
```
JmPotato, line 110: Do we still need this type?

BornChanger: Better to keep it there in case some customers wanna distinct different workload rus if the overhead is minor.

proto/resource_manager.proto
```diff
@@ -119,24 +118,25 @@ enum RawResourceType {
...
 message Consumption {
-  double r_r_u = 1;
-  double w_r_u = 2;
-  double read_bytes = 3;
-  double write_bytes = 4;
-  double total_cpu_time_ms = 5;
-  double sql_layer_cpu_time_ms = 6;
-  double kv_read_rpc_count = 7;
-  double kv_write_rpc_count = 8;
+  double r_u = 1;
```
glorv, line 121: Since r_u is always equal to r_r_u + w_r_u, do we still need to add this field?

proto/resource_manager.proto
CabinfeverB: > Better to keep it there in case some customers wanna distinct different workload rus if the overhead is minor.

Customers can distinct different workload rus by `Consumption`. In fact the type only indicates that it will be used in token bucket.
I want keep RequestUnit mode and RawResource mode same in `TokenBucketRequest`

JmPotato (approved):

BornChanger (approved):

PR #1051, opened 2023-01-31
fix item value type

HuSharp:
As https://github.com/tikv/tikv/issues/14104, we need to provide new filed to support item value as bytes


disksing (approved):

proto/pdpb.proto
```diff
@@ -139,9 +139,13 @@ enum EventType {
 
 message GlobalConfigItem {
     string name = 1;
+    // this field 'value' is replaced by the field 'value_payload'.
     string value = 2;
     Error error = 3;
     EventType kind = 4;
+    // Since item needs to support different struct types, 
+    // it should be set bytes instead of string
+    bytes value_payload = 5;
```
glorv, line 148:
```suggestion
    bytes payload = 5;
```

glorv (approved):
LGTM

PR #1058, opened 2023-02-20
add force update

HuSharp:
In order not to block local branch development, when meet break compatibility will force to update `proto.lock`.

Before this pr, when developing locally, because the development is not done overnight, guys may update the serial number. But because guys do not know the `proto mechanism` too well, will be confused to encounter the problem of proto lock update. But in fact, the main role of `proto lock` is to prompt the role of checking.
So I modified it to force the update and only make it a hint.

when `make check`, will:
image

Instead of reporting errors to organize the development process


HuSharp:
@JmPotato @binshi-bing PTAL! thx!

JmPotato (commented):
LGTM

scripts/check.sh
```diff
@@ -26,9 +26,10 @@ check-protos-compatible() {
 
     if protolock status -lockdir=scripts -protoroot=proto; then
         protolock commit -lockdir=scripts -protoroot=proto
-        return 0
     else
-        return 1
+        echo "Meet break compatibility problem, please check the code."
```
binshi-bing, line 30: Is it possible to check out the latest version of proto.lock before all the local commits to compare with the current proto.lock to find compatibility issue? If we can't, the change looks good to me.

HuSharp: After looking at the `protolock` source code we used, the code uses absolute paths and cannot set the path as github master. :(
And i add `git diff` check to avoid not adding `proto.lock` to git.

binshi-bing (approved):
LGTM

HuSharp:
@glorv @JmPotato @nolouch can u take a look? thx!

scripts/check.sh
rleungx: How about only enabling it in the CI environment?

HuSharp: If we put it in the ci, the development won't be able to commit the protolock every time when he makes a change.
Do you mean to make the robot implement add protolock Commit? I am not sure does it have this mechanism.

scripts/check.sh
```diff
@@ -26,10 +26,14 @@ check-protos-compatible() {
...
+        # In order not to block local branch development, when meet break compatibility will force to update `proto.lock`.
```
nolouch, line 31: why not the developers use a command to force updates by themselves?

HuSharp: Because developers do not know the proto mechanism too well, will be confused to encounter the problem of proto lock update.
And the main role of proto lock is for checking just like a hint,
So force update it in `make check` to decrease developers' learning cost.

Connor1996 (approved):
LGTM

PR #1060, opened 2023-02-21
Add s3 lock service

Lloyd-Pottiger:


proto/kvrpcpb.proto
```diff
@@ -1428,3 +1428,55 @@ message CompactResponse {
     bytes compacted_start_key = 3;
     bytes compacted_end_key = 4;
 }
+
+message S3LockError {
```
JaySon-Huang, line 1432: Move all these message into `disaggregated.proto`

Lloyd-Pottiger: ok

Lloyd-Pottiger: done

JaySon-Huang (approved):

PR #1061, opened 2023-02-22
Add watch RPC

rleungx:


lhy1024 (approved):

proto/pdpb.proto
```diff
@@ -113,6 +119,53 @@ message WatchGlobalConfigResponse {
...
+message GetRequest {
+    bytes key = 1;
+    bytes range_end = 2;
+    int64 limit = 3;
+    int64 revision = 4;
+}
+
+// copied part of https://github.com/etcd-io/etcd/blob/7dfd29b0cc7ce25337276dce646ca2a65aa44b4d/api/etcdserverpb/rpc.proto
+message GetResponse {
+    ResponseHeader header = 1;
```
YuJuncen, line 147: I think the response header should be a trimmed of this:
https://github.com/etcd-io/etcd/blob/3cf2e79e16a7a0e08de0ac168c829b4df20ae644/api/etcdserverpb/rpc.proto#L391-L406

(At least I think we need to keep the `revision` field of this...)

rleungx: Done. Add the revision in the corresponding response.

proto/pdpb.proto
```diff
@@ -103,8 +103,9 @@ service PD {
 }
 
 message WatchGlobalConfigRequest {
-    string config_path = 1;
-    int64 revision = 2;
+    string source = 1;
```
glorv, line 106: this is change will break the backward compatibility

rleungx: It is changed by accident.

rleungx: fixed

rleungx:
BTW, should we name it storagepb instead of etcdpb? @binshi-bing @nolouch

binshi-bing:
> BTW, should we name it storagepb instead of etcdpb? @binshi-bing @nolouch

+1 for storagepb or kvstoragepb

binshi-bing (approved):
LGTM

proto/storagepb.proto
```diff
@@ -0,0 +1,136 @@
...
+
+// copied part of https://github.com/etcd-io/etcd/blob/7dfd29b0cc7ce25337276dce646ca2a65aa44b4d/api/etcdserverpb/rpc.proto
+message WatchResponse {
+    ResponseHeader header = 1;
+    int64 compact_revision = 2;
+    repeated Event events = 3;
+}
+
+// copied part of https://github.com/etcd-io/etcd/blob/7dfd29b0cc7ce25337276dce646ca2a65aa44b4d/api/etcdserverpb/rpc.proto
+message GetRequest {
```
HuSharp, line 68: Do you think we need to add more comments to indicate that the `GetRequest` here corresponds to etcd's `RangeRequest`?

rleungx: Both are ok if we are going to use another kind of storage, Get sounds more general.

nolouch:
> > BTW, should we name it storagepb instead of etcdpb? @binshi-bing @nolouch
> 
> +1 for storagepb or kvstoragepb

I think storagepb or kvstoragepb maybe too general. we have TiKV and Tiflash storage.

YuJuncen:
Perhaps we can delete the generated `etcdpb.pb.go` for now?
BTW, I'm wondering will changing the service name make implement the client harder?
Rest LGTM.

rleungx:
> Perhaps we can delete the generated `etcdpb.pb.go` for now? BTW, I'm wondering will changing the service name make implement the client harder? Rest LGTM.

Yes, I noticed that and did a cleanup.

rleungx:
> > > BTW, should we name it storagepb instead of etcdpb? @binshi-bing @nolouch
> > 
> > 
> > +1 for storagepb or kvstoragepb
> 
> I think `storagepb` or `kvstoragepb` maybe too general. we already have TiKV and TiFlash storage, it's confusing. I prefer `etcdpb` or `metastoragepb`.

change to `metastoragepb`

rleungx:
@YuJuncen @nolouch comment addressed, PTAL

proto/storagepb.proto
rleungx: add a comment in RPC definition

YuJuncen (approved):
LGTM

nolouch (approved):
lgtm

PR #1066, opened 2023-02-28
importer: add missing region id for compact request

BusyJay:


tabokie (approved):
title should be "missing context"?

lance6716 (approved):

PR #1072, opened 2023-03-09
Introduce incr snap

BusyJay:
This is the refined snapshot protocol for v2.


proto/raft_serverpb.proto
```diff
@@ -60,6 +60,53 @@ message SnapshotChunk {
 
 message Done {}
 
+message TabletSnapshotFileMeta {
+    uint64 file_size = 1;
+    string file_name = 2;
+    // Some block data.
+    bytes head_chunk = 3;
```
tonyxuqqi, line 67: is it better to separate the checksum for quick compare?

proto/raft_serverpb.proto
```diff
@@ -60,6 +60,53 @@ message SnapshotChunk {
...
+    // trailing data including checksum.
+    bytes trailing_chunk = 4;
+}
+
+message TabletSnapshotPreview {
+    repeated TabletSnapshotFileMeta meta = 1;
```
tonyxuqqi, line 73: > meta

metas

proto/raft_serverpb.proto
```diff
@@ -60,6 +60,53 @@ message SnapshotChunk {
...
+    // There may be too many metas, use a flag to indicate all metas
+    // are sent.
+    bool end = 2;
+}
+
+message TabletSnapshotFileChunk {
+    uint64 file_size = 1;
+    string file_name = 2;
+    // data does not include the first chunk sent in `TabletSnapshotPreview`.
+    bytes data = 3;
```
tonyxuqqi, line 83: Is the data has self-sufficient data integrity mechanism? If not, I think it's better to include the checksum inside this struct. Relying the caller to store checksum somewhere is not as handy.

proto/raft_serverpb.proto
```diff
@@ -60,6 +60,53 @@ message SnapshotChunk {
...
+        TabletSnapshotHead head = 1;
+        TabletSnapshotPreview preview = 2;
+        TabletSnapshotFileChunk chunk = 3;
+        // Server side stream can finish without error, tablet snapshot doesn't
+        // have checksum, so we use an explicit message to mark the end of stream.
+        Done done = 4;
+    }
+}
+
+message AcceptedFiles {
```
tonyxuqqi, line 102: it might be better to rename as AcceptedSnapshotFiles.

proto/raft_serverpb.proto
```diff
@@ -60,6 +60,53 @@ message SnapshotChunk {
...
+    repeated string file_name = 1;
+}
```
tabokie, line 104: Seems redundant to have a dedicated message

tabokie (approved):

proto/raft_serverpb.proto
BusyJay: That will require parsing the SST file.

proto/raft_serverpb.proto
BusyJay: I do that for extensibility in case we want to change response to oneof style just like request in the future.

proto/raft_serverpb.proto
BusyJay: Now a checksum will be computed for the whole snapshot and included in the `TabletSnapshotEnd` message.

tonyxuqqi (approved):

proto/raft_serverpb.proto
```diff
@@ -60,6 +60,59 @@ message SnapshotChunk {
...
+
+message TabletSnapshotHead {
+    RaftMessage message = 1;
+    bool use_cache = 2;
+}
+
+message TabletSnapshotEnd {
+    // Checksum of all data sent in `TabletSnapshotFileChunk.data`.
+    uint64 checksum = 1;
```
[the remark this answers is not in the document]
BusyJay: There still be chance in software bugs or hardware errors. We have observed message corruption in the application layer (like raft message) in practice. Note the overhead is small. According to https://github.com/tikv/crc64fast, it can process more than 28GiB per second. The major overhead is from IO.

PR #1073, opened 2023-03-09
Add start_ts field into kvrpcpb.Context

JmPotato:
Ref https://github.com/pingcap/tidb/issues/41972, https://github.com/tikv/client-go/pull/732.


scripts/proto.lock
```diff
@@ -10263,6 +10268,11 @@
                 "name": "logical_table_id",
                 "type": "int64"
               },
+              {
+                "id": 7,
+                "name": "api_version",
```
Connor1996, line 10273: why hasn't it been updated before?

JmPotato: I assume that the previous guy didn't submit the `proto.lock` file 🤔.

proto/kvrpcpb.proto
```diff
@@ -845,6 +845,9 @@ message Context {
     // error before processing if the estimated waiting duration exceeds the threshold.
     uint32 busy_threshold_ms = 27;
 
+    // The `start_ts` of the transaction that the request belongs to.
+    uint64 start_ts = 28;
```
Connor1996, line 849: I notice a lot of requests already have a `start_version` field. Why bother add one more in context

JmPotato: After digging into the code, I think adding a `start_ts` field is the simplest way to get it passed into the place where the resource control works at.

Please take a look at https://github.com/tikv/client-go/pull/732/files#diff-34812587e42268a5d84933be286601fc58fdff17ab336b09ba75455ffe5ed447R47. `tikvrpc.Request` contains `kvrpcpb.Context`.

scripts/proto.lock
HuSharp: @Connor1996 can u take a look at https://github.com/pingcap/kvproto/pull/1058? :)

proto/kvrpcpb.proto
Connor1996: But you can get the start ts from `tikvrpc.Request` by a type matching

nolouch: BTW, The PointGet may use `max_ts`? Will this affect the result?

JmPotato: I double-checked the code of TiDB and think it's not a problem. Here is the reason: https://github.com/tikv/client-go/pull/732#discussion_r1135780987.

PR #1077, opened 2023-03-14
support service mode switching for tso

binshi-bing:
Add GetClusterInfo rpc to PD
Add tsopb.Participant as the value type of tso primary


proto/pdpb.proto
```diff
@@ -389,6 +393,22 @@ message GetMembersResponse {
     map<string, Member> tso_allocator_leaders = 5;
 }
 
+message GetClusterInfoRequest {
+    ResponseHeader header = 1;
+}
+
+enum ServiceMode {
+    UNKNOWN_SVC_MODE = 0;
```
rleungx, line 401: Do we need unknown?

binshi-bing: Yes, it needs to an initial value at startup.
1. If I initialize it with PD module, it actually introduce more complexity at startup.
2. We might need to set UNKNOWN_SVC_MODE by server side if the service is in switching, e.g., half servers running in pd mode; and half servers running in api mode, so we might need one more state.

rleungx (approved):

PR #1093, opened 2023-03-28
update go version to `1.18.2` in docker

Lloyd-Pottiger:
I meet the following error: 

```
Cannot load io/fs: malformed module path "io/fs": missing dot in first path element
```

Due to the version of go is `1.13` which is two low, update to `1.18.2`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,13 @@ RUN apt-get update --yes \
  && apt-get install --yes curl unzip \
  && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


HuSharp:
Can you explain why it is specific `1.18.2`?

Lloyd-Pottiger:
> 

https://github.com/pingcap/kvproto/blob/d7e4db66a63e2f5866f0a452bcddc842fec531d0/go.mod#L23 just the same as the main version go this repo. 1.18.0-1.18.10 are all acceptable.

disksing (approved):

PR #1094, opened 2023-03-28
Support return TiFlash system table by grpc

Lloyd-Pottiger:
Signed-off-by: Lloyd-Pottiger <yan1579196623@gmail.com>


disksing:
generally LGTM. Is it ready to merge or need someone else to review?

scripts/check.sh
```diff
@@ -34,6 +34,11 @@ check-protos-compatible() {
     # If the output message is encountered, please add proto.lock to git as well.
     git diff scripts/proto.lock | cat
     git diff --quiet scripts/proto.lock
+    if [ $? -ne 0 ]; then
```
HuSharp, line 37: As described in this pr https://github.com/pingcap/kvproto/pull/1058
I modified it to force the update and only make it like a hint, I think it's better to keep it like a hint. Rather than return error to confuse developers and it will also increase developers' learning costs.

hongyunyan (approved):
LGTM

PR #1103, opened 2023-04-19
Add `source_stmt` in `Context` for test logging.

ekexium:
ref tikv/tikv#14474


proto/kvrpcpb.proto
```diff
@@ -916,6 +920,11 @@ enum APIVersion {
     V2 = 2;
 }
 
+message TidbSource {
```
cfzjywxk, line 923: How about `QueryCtx` or `StmtCtx`? Our goal is to establish a connection between the execution flow of transactions or statements.
`connection_id` may not be globally unique, to clarify which query the request belongs to something like `stmt_id` or `uuid` is needed but it looks a bit complex.

proto/kvrpcpb.proto
```diff
@@ -862,6 +862,10 @@ message Context {
     // The keyspace that the request is sent to.
     // NOTE: This field is only meaningful while the api_version is V2.
     uint32 keyspace_id = 32;
+    
+    // It tells us where the request comes from in TiDB. If it isn't from TiDB, leave it blank.
+    // This is for tests only and thus can be safely changed/removed without affecting compatibility.
+    SourceStmt source_stmt = 33;
```
zyguan, line 868: If this field is only used for logging with context info, I think a formated string is enough (just like `request_source`).

ekexium: IMO string is less structured and less flexible.

disksing:
@cfzjywxk @zyguan Please take another look.

proto/kvrpcpb.proto
```diff
@@ -938,6 +942,12 @@ enum APIVersion {
     V2 = 2;
 }
 
+message SourceStmt {
+    uint64 start_ts = 1;
+    uint64 connection_id = 2;
+    uint64 stmt_id = 3;
+}
```
cfzjywxk, line 949: Since the `connection_id` can be identical across different TiDB servers, it may be necessary to use a unique identifier such as `uuid` to indicate which statement and instance this query originates from.
Besides, how should the `stmt_id` be generated?

zyguan (approved):
LGTM, is the source_stmt field only set for testing?

BTW, source_stmt.start_ts seems to be a little redundant since most of rpc requests already have a field that indicate to start-ts.

proto/kvrpcpb.proto
ekexium: > how should the stmt_id be generated

Each connection generates its own stmt_id. 
Merging them together into a UUID can make it harder for humans to read. Since it's only used in tests, the performance gain doesn't matter a lot.

ekexium:
We can wait until the corresponding TiKV PR completes its review before merging this PR

PR #1105, opened 2023-04-21
Add a flag to CheckTxnStatusRequest to tell TiKV to check whether the lock on the key (if any) is a real primary lock

MyonKeminta:
Ref: https://github.com/pingcap/tidb/issues/42937


zyguan (approved):

proto/kvrpcpb.proto
```diff
@@ -309,6 +309,12 @@ message CheckTxnStatusRequest {
     // lock, the transaction status could not be decided if the primary lock is pessimistic too and
     // it's still uncertain.
     bool resolving_pessimistic_lock = 8;
+    // Whether it's needed to check if the lock on the key (if any) is the primary lock.
+    // This is for handling some corner cases when pessimistic transactions changes its primary
```
ekexium, line 313:
```suggestion
    // This is for handling some corner cases when a pessimistic transaction changes its primary
```

ekexium (approved):

PR #1111, opened 2023-05-08
introduce keyspace safepoint interface

ystaticy:


proto/pdpb.proto
```diff
@@ -815,6 +823,70 @@ message UpdateServiceGCSafePointResponse {
...
+
+message WatchGCSafePointV2Request {
+    RequestHeader header = 1;
+}
+
+// SafePointEvent is for the rpc WatchGCSafePointV2.
+message SafePointEvent {
+    uint32 keyspace_id = 1;
+    uint64 safe_point = 2;
+    EventType type = 3;
```
MyonKeminta, line 846: What's the semantics of the `EventType` used here?

proto/pdpb.proto
```diff
@@ -815,6 +823,70 @@ message UpdateServiceGCSafePointResponse {
...
+    // This should be set by component that may crash unexpectedly so that it doesn't block
+    // cluster garbage collection.
+    int64 ttl = 5;
+}
+
+message UpdateServiceSafePointV2Response {
+    ResponseHeader header = 1;
+
+    bytes service_id = 2;
+    int64 TTL = 3;
```
MyonKeminta, line 886: Use lower case here

proto/pdpb.proto
ystaticy: When using the etcd watch interface, it represents the put/delete operation on the etcd path.

MyonKeminta (approved):

rleungx (approved):

PR #1128, opened 2023-06-12
Add resolved ts traffic optimizations

you06:
implement tikv/rfcs#108


disksing:
@you06  need run make to generate the lock file.

PR #1130, opened 2023-06-12
add a new RPC EventFeedV2 for cdc

hicqu:
Currently there is already a RPC for cdc `EventFeed`, which has such agreements:

* In one stream a region can't be subscribed more than 1 times.

This makes TiCDC tables can't share one stream, which may cause resource wasting.

To avoid this problem, this PR adds a new RPC `EventFeedV2`.

### Why must a new RPC?

Rolling upgrade behavior will be changed soon: TiCDC will be upgraded before TiKV.
So to avoid mistakes, we need a new RPC to report "unimplemented" to clients in this case.

### What if more changes in future? Will we need more RPCs?

No. We can add a `requested features` check in `EventFeedV2` server side. All other changes can be covered by that check.


hi-rustin (approved):

proto/cdcpb.proto
```diff
@@ -166,10 +174,17 @@ message ChangeDataRequest {
...
 }
 
 service ChangeData {
     rpc EventFeed(stream ChangeDataRequest) returns(stream ChangeDataEvent);
+
+    // EventFeedV2 is like EventFeed, with some new changes:
+    //  * multiplexing, which means one region can be subscribed multiple times
+    //    with different `request_id` in one RPC stream. And, it can be
+    //    deregistered with a `Deregister` message with a given `request_id`.
+    rpc EventFeedV2(stream ChangeDataRequest) returns (stream  ChangeDataEvent);
```
overvenus, line 189: The RPC signature is identical to `EventFeed`,  can we extend `EventFeed` instead?

TiKV CDC component has feature gate which we can enable or disable multiplexing.

hicqu: Reuse `EventFeed` can lead mistakes if TiCDC version is higher than TiKV.
So `EventFeedV2` is required for this case. And I have added version and requested feature checks in `EventFeedV2`.
Please check [here](https://github.com/tikv/tikv/pull/14923/files#diff-ae1198b4c32440ad526f4bfc1ea4ce851d7d46aa9ba8b9000c37a42d770b64b4R411).

hicqu: /cc @hi-rustin @asddongmen

overvenus: > Reuse EventFeed can lead mistakes if TiCDC version is higher than TiKV.

Why not let TiCDC check TiKV version before sending requests.

hicqu: If you mean we ask this infomation from PD, I think it's more complex than using another RPC. We must handle this case: PD tells us `tikv is v 6.6.6` but when establishing the connection, tikv has been restarted in `v6.6.7`.

overvenus: > If you mean we ask this infomation from PD, I think it's more complex than using another RPC. 

Checking PD is like 5-6 lines of code, but making a new RPC it duplicates at least 100 lines of code (see event_feed implementation in tikv).
I don't think the new RPC is necessary, given the fact that they has the same signature and it adds extra maintenance burden.

> PD tells us tikv is v 6.6.6 but when establishing the connection, tikv has been restarted in v6.6.7

It's rare case, because TiCDC is upgraded after TiKV. And even if it happens it won't cause issue, because new tikv must be compatible with old protocol.

hicqu: The behavior will be changed soon. TiCDC will be upgraded before TiKV later.

overvenus: It's still the rare case, because tikv updates version in pd before starts serving, very little chance of triggering such issue, and even if it happens it won't cause issue, because new tikv must be compatible with old protocol.

Also if you really care about this rare issue, you can probe tikv by sending two requests for DDL region (which is needed for changefeeds that have more than 2 tables).

overvenus:
Why merge? There is one LGTM.

PR #1134, opened 2023-06-16
add context for br and import_sst

nolouch:


proto/import_sstpb.proto
```diff
@@ -281,6 +281,7 @@ message RawWriteRequest {
         SSTMeta meta = 1;
         RawWriteBatch batch = 2;
     }
+    kvrpcpb.Context context = 3;
```
glorv, line 284: should also add for WriteRequest

glorv (approved):

PR #1138, opened 2023-06-27
switch mode request

SpadeA-Tang:


proto/import_sstpb.proto
```diff
@@ -70,8 +70,15 @@ enum SwitchMode {
...
 
+// [start, end)
+message KeyRange {
+    bytes start = 1;
+    bytes end = 2;
+}
+
 message SwitchModeRequest {
     SwitchMode mode = 1;
+    KeyRange range = 2;
```
overvenus, line 81: How about using `kvrpcpb.KeyRange`?

SpadeA-Tang: Done

overvenus (approved):

PR #1141, opened 2023-06-29
[6.1] Cherry-pick changes about the fix of pessimistic transactions with primary changed

MyonKeminta:
ref: https://github.com/pingcap/tidb/issues/42937

Cherry-picks https://github.com/pingcap/kvproto/pull/1105 https://github.com/pingcap/kvproto/pull/1107 to release-6.5 branch


ekexium (approved):

cfzjywxk (approved):

MyonKeminta:
Sorry. I noticed that some released versions with larger minor version number than 6.1 (like 6.5) doesn't contain this fix. Fixing this in 6.1 branch might cause problem if there are any user trying to upgrade to older version in newer branch (such as upgrading from 6.1.7 to 6.5.0). For this reason, whether this bug should be fixed in 6.1 branch needs discussion.
cc @cfzjywxk

cfzjywxk:
Got it,I think to keep compatibility we could skip the fix on v6.1.

PR #1142, opened 2023-06-30
kvproto: add request source v2

nolouch:
TiKV will sue request source to juge background job, so we add enum type for it.
```
- add request_source_v2
- remove the `is_background` item (There is  no dependence here, so we can remove it directly)
```


proto/kvrpcpb.proto
```diff
@@ -865,11 +865,32 @@ message Context {
...
     // NOTE: This field is only meaningful while the api_version is V2.
     uint32 keyspace_id = 32;
 }
 
+enum RequestSourceType {
+    Default = 0;
+    Lightning = 1;
+    Dumpling = 2;
+    BR = 3;
+    Background = 4;
```
glorv, line 882:
```suggestion
    Background = 4;
    DDL = 5;
    TTL = 6;
    Stats = 7; // auto background tasks such as analyze.
    GC = 8;  // do we need this?
```
Not sure if we also

nolouch:
we decide to reuse `request_source` first.

PR #1143, opened 2023-06-30
restore: support file-copy restore

3pointer:
ref https://github.com/pingcap/tidb/issues/44942


proto/import_sstpb.proto
```diff
@@ -153,6 +153,10 @@ message CompactResponse {
 }
 
 message DownloadRequest {
+    // Map represents the map of <name, SSTMeta>.
+    // We'll generate all SSTMeta into one SST File.
+    map<string, SSTMeta> ssts = 1;
```
disksing, line 158: is field id 1 never used before?

3pointer: It seems index starts from 2, and 1 was never used.

disksing (approved):

PR #1145, opened 2023-07-03
Add commit index to snapshot meta

overvenus:
Include leader's latest commit index in snpashot so that tiflash knows when it can get a consistent (no default_not_found error) tablet snapshot.

See https://github.com/pingcap/tiflash/issues/7568


CalvinNeo (approved):
lgtm

tonyxuqqi (approved):

PR #1147, opened 2023-07-05
add `gather_id` to `DisaggTaskMeta`

windtalker:


proto/disaggregated.proto
```diff
@@ -65,14 +65,15 @@ message GetDisaggConfigResponse {
...
-    // query_ts + server_id + local_query_id to represent a global unique query.
+    // gather_id + query_ts + server_id + local_query_id to represent a global unique query.
     uint64 query_ts = 2; // timestamp when start to execute query, used for TiFlash miniTSO schedule.
     uint64 server_id = 3; // TiDB server id
     uint64 local_query_id = 4; // unique local query_id if tidb don't restart.
     int64 task_id = 5; // if task id is -1 , it indicates a tidb task.
     string executor_id = 6; // the exectuor id
     uint32 keyspace_id = 7; // keyspace id of the request
     kvrpcpb.APIVersion api_version = 8; // API version of the request
+    int64 gather_id = 9;
```
JaySon-Huang, line 76: Better put `gather_id` next to `local_query_id`.

And we should also update the comment in "mpp.proto"
https://github.com/pingcap/kvproto/blob/cd6769a15c69b48fd65ecb6d813b861156be4d61/proto/mpp.proto#L23

windtalker: I'm not sure if protobuf support insert a field in the middle of original fields

JaySon-Huang: you can put it in the middle but keep the `gather_id = 9;`. The protobuf only identified a field by the `id`

windtalker: OK, updated.

JaySon-Huang (approved):
LGTM

PR #1148, opened 2023-07-05
support setting multiple ranges in switch mode request 

SpadeA-Tang:


lance6716 (approved):

proto/import_sstpb.proto
```diff
@@ -72,7 +72,7 @@ enum SwitchMode {
 
 message SwitchModeRequest {
     SwitchMode mode = 1;
-    Range range = 2;
+    repeated Range range = 2;
```
lance6716, line 75:
```suggestion
    repeated Range ranges = 2;
```

PR #1152, opened 2023-07-10
extra_msg: add bucket version message to notify follower refresh bucket.

bufferflies:
In past, follower read response doesn't carry bucket version, so the client don't know whether to refresh the buckets key range.


proto/raft_serverpb.proto
```diff
@@ -246,12 +246,17 @@ enum ExtraMessageType {
...
     MsgGcPeerResponse = 12;
     MsgFlushMemtable = 13;
+    MsgRefreshBuckets = 14;
 }
 
 message FlushMemtable {
     uint64 region_id = 1;
 }
 
+message RefreshBuckets{
```
tabokie, line 256:
```suggestion
message RefreshBuckets {
```

tabokie (approved):

proto/raft_serverpb.proto
bufferflies: fixed

PR #1157, opened 2023-07-14
remove useless is_background

nolouch:


disksing:
need to mark it as reserved?

nolouch:
@disksing It's introduced by https://github.com/pingcap/kvproto/pull/1132, and is not yet being used by other components.

disksing (approved):

PR #1160, opened 2023-07-20
resource_manager: add is_background check

HuSharp:
add is_background check for tikv upload ru


proto/resource_manager.proto
```diff
@@ -61,6 +61,8 @@ message TokenBucketsRequest {
   repeated TokenBucketRequest requests = 1;
   uint64 target_request_period_ms = 2;
   uint64 client_unique_id = 3;
+  // label background request.
+bool is_background = 4;
```
nolouch, line 65: format.

HuSharp: fixed.

nolouch (approved):

proto/resource_manager.proto
```diff
@@ -61,6 +61,8 @@ message TokenBucketsRequest {
   repeated TokenBucketRequest requests = 1;
   uint64 target_request_period_ms = 2;
   uint64 client_unique_id = 3;
+  // label background request.
+  bool is_background = 4;
```
glorv, line 65: Is it better to put this field into TokenBucketRequest? We may want to batch both background/foreground request together

nolouch: cc @HuSharp  :)

HuSharp: Okay, I opened a new pr to change it.https://github.com/pingcap/kvproto/pull/1161

PR #1162, opened 2023-07-24
Resource manager/watch type

CabinfeverB:
Add watch type `plan`. And modify last_duration_ms as int64 because of `time.Duration` is int64 and we should use negative to indicate `unlimited`.


proto/resource_manager.proto
```diff
@@ -196,6 +196,7 @@ enum RunawayAction {
 enum RunawayWatchType {
   Exact = 0;
   Similar = 1;
+  Plan = 3;
```
glorv, line 199: If I understand correctly, in the current impl, Similar actually means Plan, so after this change Similar will mean SQL digest, right? So this is a incompatible change

CabinfeverB: Yes, we should also modify the docs

glorv (approved):
LGTM

proto/resource_manager.proto
```diff
@@ -204,7 +205,7 @@ message RunawayRule {
 
 message RunawayWatch {
   // how long would the watch last
-  uint64 lasting_duration_ms = 1;
+  int64 lasting_duration_ms = 1;
```
Connor1996, line 208: why not use 0 as unlimited

CabinfeverB: In TiDB, ResourceGroupRunawaySettings is defined as follow,
```
type ResourceGroupRunawaySettings struct {
	ExecElapsedTimeMs uint64            `json:"exec_elapsed_time_ms"`
	Action            RunawayActionType `json:"action"`
	WatchType         RunawayWatchType  `json:"watch_type"`
	WatchDurationMs   int64             `json:"watch_duration_ms"`
}
```
I should use WatchDurationMs to check whether `Watch` is defined.
But I can use a pointer for `Watch`.   Which do you think is better

Connor1996: maybe add a none watch type?

glorv: +1

Connor1996 (approved):
LGTM

PR #1164, opened 2023-07-25
Cherry pick #1155 and #1158 to release-6.5

ekexium:
Compatibility:
The patch adds one RPC `get_region_read_progress` in debug service, which can only be triggered by `tikv-ctl get-region-read-progress` command. Updating 6.5.x to a newer major version without this patch simply removes the command. When using tikv-ctl on a different version of TiKV, an unimplemented error could be returned. Anyway, TiKV itself will not be impacted.


cfzjywxk:
As it's a proto change on release branch, it's needed to describe the compatibilty impact.

ekexium:
Compatibility details have been added.

cfzjywxk (approved):

PR #1165, opened 2023-07-26
Add skip store limit check for scatter region

rleungx:
ref https://github.com/tikv/pd/pull/6593


proto/pdpb.proto
```diff
@@ -800,6 +800,7 @@ message ScatterRegionRequest {
     // If regions_id is defined, the region_id would be ignored.
     repeated uint64 regions_id = 6;
     uint64 retry_limit = 7;
+    bool skip_limit_check = 8;
```
nolouch, line 803: maybe `skip_store_limit`  or `skip_store_limit_restriction`

`limit_check` may confuse with `retry_limit`

rleungx: I prefer a more general name but if we don't find one, `skip_store_limit` is ok to me.

3pointer (approved):

PR #1171, opened 2023-08-10
Add new read response for reject follower read when follower log behind of read index a lot.

bufferflies:
tikv pr：https://github.com/tikv/tikv/pull/15270
tikv issue： https://github.com/tikv/tikv/issues/15189


proto/errorpb.proto
```diff
@@ -161,6 +161,13 @@ message MismatchPeerId {
     uint64 store_peer_id = 2;
 }
 
+message FollowerNotReady{
+    uint64 region_id = 1;
+    uint64 read_index = 2;
+    uint64 apply_index = 3;
+    uint64 peer_id = 4;
```
overvenus, line 168: Are they necessary for error handling? How are they used?

bufferflies: yes, only used for debug, I will change it as reason.

PR #1175, opened 2023-08-21
support trace sql connection_id and session_alias

lcwangchao:
We need to add some sql fields to trace some user operations in full link


cfzjywxk:
Could this https://github.com/pingcap/kvproto/pull/1103 be resued? 
@ekexium PTAL

ekexium:
> Could this #1103 be resued? @ekexium PTAL

connection_id seems to have the same meaning. I believe we can use either of these structures as long as we put everything we need in it.

lcwangchao:
closed as #1103  has implemented it

PR #1177, opened 2023-08-31
Add new property to mark the initial size for the new region.

bufferflies:


proto/raft_cmdpb.proto
```diff
@@ -146,6 +146,9 @@ message SplitRequest {
     // left region use new_region_id.
     // Will be ignored in batch split, use `BatchSplitRequest::right_derive` instead.
     bool right_derive = 4 [deprecated=true];
+    // It should be false iff the region split by user key such as split table or create partion table etc, and new region's size should be zero. 
+    // It should be true iff the region's load reach the threshold such as size、keys、load check etc, and new region's size will amortize the origin region.
```
SpadeA-Tang, line 150: Maybe change to "It determines whether the split-off regions amortize the region size of the source region. In some cases, such as ..., the split-off regions won't have any data, so they should't amortize the source region's region size." ?

SpadeA-Tang (approved):

protoc-3.15.8-linux-x86_64.zip
rleungx, line 1: Please remove it

proto/raft_cmdpb.proto
bufferflies: I will change the behavior description.

protoc-3.15.8-linux-x86_64.zip
bufferflies: fixed

proto/raft_cmdpb.proto
```diff
@@ -146,6 +146,11 @@ message SplitRequest {
     // left region use new_region_id.
     // Will be ignored in batch split, use `BatchSplitRequest::right_derive` instead.
     bool right_derive = 4 [deprecated=true];
+    // It should be false iff the region split by user key such as split table or create partion table etc, 
+    // the new region's will not amortize the source region size, so it's region size is zero.
+    // It should be true iff the region's load reach the threshold such as size、keys、load check etc,
```
rleungx, line 151:
```suggestion
    // It should be true iff the region's load reaches the threshold such as size, keys, load check, etc,
```

proto/raft_cmdpb.proto
```diff
@@ -158,6 +163,11 @@ message BatchSplitRequest {
     // If true, the last region derive the origin region_id, 
     // other regions use new ids.
     bool right_derive = 2;
+    // It should be false iff the region split by user key such as split table or create partion table etc, 
+    // the new region's will not amortize the source region size, so it's region size is zero.
+    // It should be true iff the region's load reach the threshold such as size、keys、load check etc,
```
rleungx, line 168: ditto

proto/raft_cmdpb.proto
bufferflies: fixed

proto/raft_cmdpb.proto
bufferflies: fixed

rleungx (approved):

PR #1180, opened 2023-09-05
Use separate participant

rleungx:
See https://github.com/tikv/pd/pull/7032


JmPotato (approved):

proto/resource_manager.proto
```diff
@@ -221,3 +221,13 @@ message RunawaySettings {
...
 }
+
+message Participant {
+    // name is the unique name of the resource manager participant.
+    string name = 1;
+    // id is the unique id of the resource manager participant.
+    uint64 id = 2;
+    // listen_urls is the serivce endpoint list in the url format.
+    // listen_urls[0] is primary service endpoint.
+    repeated string listen_urls = 3;
```
lhy1024, line 232: repeated string？

rleungx: keep the same as the tso proto

lhy1024 (approved):

PR #1190, opened 2023-09-20
raft_serverpb: add removed records to merged record

overvenus:


proto/raft_serverpb.proto
```diff
@@ -191,6 +191,7 @@ message MergedRecord {
     uint64 source_region_id = 1;
     metapb.RegionEpoch source_epoch = 2;
     repeated metapb.Peer source_peers = 3;
+    repeated metapb.Peer source_removed_records = 9;
```
tonyxuqqi, line 194: It's better to move it to the bottom.  Otherwise next change may still use 9.

tonyxuqqi (approved):

proto/raft_serverpb.proto
tonyxuqqi: Why removed_records cannot be used for this purpose

overvenus: > Why removed_records cannot be used for this purpose

Because peers in `removed_records` belong target region, `source_removed_records` belong to source region.

SpadeA-Tang (approved):

PR #1204, opened 2023-11-03
Added a new `PrepareBackup` RPC for snapshot backup

YuJuncen:
### Q & A

Some questions you may like to ask:

#### Q: Is it possible to split this into two RPCs?

Surely we can split this into two RPCs: one for ready and one for wait applies.

This will allow us to reuse some gRPC's infras to control connections, and easier to be understood the behavior.

But I don't want to do this because:
- It seems waiting apply is meanless if we haven't disabled new commands proposing, which seems even harder to understand "why this RPC even exists?".
- Controlling many asynchronous requests seems not really easier than a sole connection with many tagged request / responses.


proto/brpb.proto
```diff
@@ -15,6 +15,30 @@ option (rustproto.lite_runtime_all) = true;
 
 option java_package = "org.tikv.kvproto";
 
+enum PrepareSnapshotBackupRequestType {
+    UpdateLease = 0;
```
Connor1996, line 19: what's this used for?

proto/brpb.proto
```diff
@@ -15,6 +15,30 @@ option (rustproto.lite_runtime_all) = true;
...
+    WaitApply = 1;
+}
+
+message PrepareSnapshotBackupRequest {
```
overvenus, line 23: Please add some comments about the message and fields.

proto/brpb.proto
YuJuncen: Updating the TTL, I will add some comments then.

proto/brpb.proto
```diff
@@ -15,6 +15,47 @@ option (rustproto.lite_runtime_all) = true;
 
 option java_package = "org.tikv.kvproto";
 
+enum PrepareSnapshotBackupRequestType {
+    // Update the lease of suspending new ingest / admin commands to be proposed.
```
Connor1996, line 19: Please describe what's the lease used for

Connor1996 (commented):
rest LGTM

PR #1209, opened 2023-11-24
add follower handle unavailable error

CabinfeverB:
ref https://github.com/tikv/pd/issues/7431


lhy1024 (approved):

CabinfeverB:
Now only GetRegion needs to support follower handle. We can reuse the error code that the region is not found. We can do this when there are other types.

PR #1218, opened 2024-01-05
Add Flush and BufferGet requests for pipelined DML

ekexium:
Ref https://github.com/tikv/tikv/issues/16291
Don't merge it until corresponding work finishes.


proto/kvrpcpb.proto
```diff
@@ -1530,3 +1530,37 @@ message TiFlashSystemTableRequest {
...
     bytes data = 1 [(gogoproto.customtype) = "github.com/pingcap/kvproto/pkg/sharedbytes.SharedBytes", (gogoproto.nullable) = false];
 }
+
+// Flush is introduced from the pipelined DML protocol. 
+// A Flush request writes some keys and values to TiKV, storing in LOCK and DEFAULT CF, just like a Prewrite request.
+message FlushRequest {
+    Context context = 1;
+    repeated Mutation mutations = 2;
+    bytes primary_key = 3;
+    uint64 start_ts = 4;
```
you06, line 1540: Shall we name the fields same with the prewrite request?

https://github.com/pingcap/kvproto/blob/932639606bcf3db9676627d55430e1dd10670570/proto/kvrpcpb.proto#L101-L106

ekexium: I'd rather rename the fields in PrewriteRequest. `start_ts` and `primary_key` are more commonly used throughout TiDB and TiKV codebase. What do you think?

you06: Yes, because we also have `for_update_ts` and `min_commit_ts`, `start_ts` is a better option.
`primary_lock` only makes sense in prewrite request, `prewrite_key` is easier to understood.

Let's leave the field names in `PrewriteRequest` as a code debt 🙈.

proto/kvrpcpb.proto
```diff
@@ -1530,3 +1530,37 @@ message TiFlashSystemTableRequest {
...
+    uint64 min_commit_ts = 5;
+    // generation of the flush request. It is a monotonically increasing number in each transaction.
+    uint64 geneartion = 6;
```
you06, line 1543: Need to run `make go` to update generation field to `kvrpcpb.pb.go`

you06:
```suggestion
    uint64 generation = 6;
```
And a typo

you06:
Need to rebase master, I cannot build the latest TiDB with this PR because some protocols are missing.

proto/kvrpcpb.proto
```diff
@@ -1534,3 +1534,39 @@ message TiFlashSystemTableRequest {
...
+    uint64 version = 3;
+}
+
+message BufferBatchGetResponse {
+    errorpb.Error region_error = 1;
+    KeyError error = 2;
+    repeated KvPair pairs = 3;
+    // Time and scan details when processing the request.
+    ExecDetailsV2 exec_details_v2 = 4;
+}
```
you06, line 1572: Leave an empty line in the end.

you06 (approved):
rest LGTM

proto/kvrpcpb.proto
```diff
@@ -1534,3 +1534,39 @@ message TiFlashSystemTableRequest {
...
+    repeated KeyError errors = 2;
+    ExecDetailsV2 exec_details_v2 = 3;
+}
+
+// BufferBatchGet is introduced from the pipelined DML protocol.
+// It is similar to a BatchGet request, except that it can only read the data that has been flushed by itself.
+message BufferBatchGetRequest {
+    Context context = 1;
+    repeated bytes keys = 2;
+    uint64 version = 3;
```
you06, line 1563: Shall we add a variable to control whether enable snapshot read after buffer read miss? This may save one RPC.

cfzjywxk: A similar question, if it's designed to support both buffer-read and snapshot read, or do we need another interface to implement "UnionStoreBatchGet" at tikv side?

ekexium: > Shall we add a variable to control whether enable snapshot read after buffer read miss? This may save one RPC.

We can do it later when we are about to implement this. We don't have to finalize all details in one PR.

>A similar question, if it's designed to support both buffer-read and snapshot read, or do we need another interface to implement "UnionStoreBatchGet" at tikv side?

`BufferBatchGet` only serves buffer read. Its semantics is "reading the content offloaded to TiKV in the membuffer"
Snapshot read does not change.
Something like `UnionStoreBatchGet` is not needed, as `unionstore.get` is simply implemented as "buffer.get, if not found, try snapshot.get". This implementation does not have to change at all. 
The only thing we need to change is how we do buffer read. Does that sound straightforward and reasonable?
```
Previous:
membuffer.get = memdb.get
snapshot.get = tikv.get
unionstore.get = membuffer.get +? snapshot.get = memdb.get +? tikv.get

Now:
membuffer.get = mutable_memdb.get +? immutable_memdb.get +? tikv.buffer_batch_get
snapshot.get = tikv.get
unionstore.get = membuffer.get +? snapshot.get = mutable_memdb.get +? immutable_memdb.get +? tikv.buffer_batch_get +? tikv.get
```

cfzjywxk:
Another way for reading a transaction's own uncommitted data in tikv:

kv-client: For pipelined-dml read, adding a flag in the existing `BatchGet` request, it indicates whether the current read is `membuf read in tikv` or `union store read in tikv`.

tikv: reuse the same `BatchGet` interface.

storage: implement the `buffer_batch_get` and `union_store_batch_get` methods.

So we could avoid introducing the `BufferGet` interface and do `union store get` within one RPC request.

ekexium:
> Another way for reading a transaction's own uncommitted data in tikv:
> 
> * kv-client: For pipelined-dml read, adding a flag in the existing `BatchGet` request, it indicates whether the current read is `membuf read in tikv` or `union store read in tikv`.
> * tikv: reuse the same `BatchGet` interface.
>   
>   * storage: implement the `buffer_batch_get` and `union_store_batch_get` methods.
> 
> So we could avoid introducing the `BufferGet` interface and do `union store get` within one RPC request.

So the only difference is in the grpc level, whether to use 3 different RPCs for snapshot read, buffer read, and union read; or use one RPC for all of them and use flags to differentiate.

Seems a trade-off between interface simplicity and clarity. Imagine when you see a batch_get from logs or metrics, we would have no idea what it is reading.

cfzjywxk:
We essentially need to support one RPC to complete the membuf read + snapshot read, regardless of whether or not a separate KV interface is split out. This scalability needs to be reserved. 
It doesn't seem to introduce too much protocol complexity to separately split out `BufferBatchGetRequest`, although it introduces a new interface. However, I don't think we need to introduce another `UnionStoreBatchGetRequest` interface.

cfzjywxk:
@MyonKeminta PTAL

MyonKeminta:
> It doesn't seem to introduce too much protocol complexity to separately split out BufferBatchGetRequest, although it introduces a new interface. However, I don't think we need to introduce another UnionStoreBatchGetRequest interface.

I agree with this. We can regard it as an RPC interface that supports *reading with the buffer*, including buffer reading only and mixed (union) reading.

But my problem is whether the word "buffer" is proper here...

ekexium:
@MyonKeminta any suggestions on names?

MyonKeminta (approved):
How about using the word "staged"?
..And maybe i'd better approve this PR for now as I don't have any better idea...

cfzjywxk (approved):
LGTM.

PR #1232, opened 2024-02-29
kv: add cluster id in batch cmd

cfzjywxk:
Ref: https://github.com/tikv/tikv/issues/16545

Add the `cluster_id` field in the `BatchCommandRequest`.


zyguan (approved):
LGTM

BTW, do we also check individual requests in batch?

cfzjywxk:
> LGTM
> 
> BTW, do we also check individual requests in batch?

I think we just check the batch.

MyonKeminta (approved):
It looks nice to me to add this and check both, but actually I think it's enough if we do the check in the common path of batched and non-batched single requests.

PR #1240, opened 2024-04-11
raft_serverpb: add an extra `ExtraMessage::MsgAckCommittedIndex` to get committed index.

LykxSassinator:
### Description
Add an extra `MsgAckCommittedIndex` message to let `peer` have the ability of fetching the committed index from the leader.


proto/raft_serverpb.proto
```diff
@@ -252,6 +252,7 @@ enum ExtraMessageType {
     MsgGcPeerResponse = 12;
     MsgFlushMemtable = 13;
     MsgRefreshBuckets = 14;
+    MsgAckCommittedIndex = 15;
```
glorv, line 255: I think use two message types for the request and resp is better.

SpadeA-Tang: +1

proto/raft_serverpb.proto
```diff
@@ -252,6 +252,8 @@ enum ExtraMessageType {
     MsgGcPeerResponse = 12;
     MsgFlushMemtable = 13;
     MsgRefreshBuckets = 14;
+    MsgAckCommittedIndexRequest = 15;
```
SpadeA-Tang, line 255: Ack is more like that someone acknowledgment something. May rename to Fetch/Ask or something?

glorv (approved):
LGTM

PR #1242, opened 2024-04-17
support pipeline dml protocol in cdc

hicqu:
This PR adds a `generation` field to `cdc.Event.Row`.


proto/cdcpb.proto
```diff
@@ -75,12 +75,16 @@ message Event {
         uint64 commit_ts = 2;
         LogType type = 3;
 
+	// generation is for pipelined DML protocol. See kvrpcpb.FlushRequest.generation.
+	uint64 generation = 10;
```
YuJuncen, line 79: Maybe fix the indent?

YuJuncen (approved):

proto/cdcpb.proto
hicqu: Indent is correct. The line is just wrapped.

flowbehappy (approved):

proto/cdcpb.proto
```diff
@@ -75,12 +75,16 @@ message Event {
         uint64 commit_ts = 2;
         LogType type = 3;
 
+        // generation is for pipelined DML protocol. See kvrpcpb.FlushRequest.generation.
+        uint64 generation = 10;
+
```
wuhuizuo, line 80: should we put it after the 9th field?

hicqu: Not necessary. If other guys add more fileds with same id, proto compiler will report it. So we can always group fields by meanning.

PR #1251, opened 2024-06-14
pdpb: support scan regions for range batch

you06:
ref pingcap/tidb#53850

Add an interface which batch scan regions for key ranges.

The test result:

1. split table into 1000 regions by `SPLIT TABLE sbtest1 BETWEEN (0) AND (1000000) REGIONS 1000;`
2. wait until region cache expired.
3. run a simple query `explain analyze select * from sbtest1 limit 1;`

- Without this interface: `...build_task_duration: 81.5ms...`, statmenet execution takes 0.11 sec
- With this interface: `...build_task_duration: 2.7ms...`, statement execution takes 0.02 sec


JmPotato (approved):

HuSharp (approved):

proto/pdpb.proto
```diff
@@ -53,6 +53,8 @@ service PD {
 
     rpc ScanRegions(ScanRegionsRequest) returns (ScanRegionsResponse) {}
 
+    rpc BatchScanRegions(BatchScanRegionsRequest) returns (BatchScanRegionsResponse) {}
```
rleungx, line 56: Will it replace ScanRegions?

you06: Yes, I think `ScanRegions` can be deprecated then.

rleungx: Then how about adding a comment for it?

you06: Done.

rleungx (approved):

cfzjywxk (approved):

PR #1252, opened 2024-07-09
pdpb: Add err type and input param to BatchScanRegions

okJiang:
ref https://github.com/tikv/pd/issues/8358


pkg/raft_serverpb/raft_serverpb.pb.go
okJiang, line 1: Automatically generated

proto/pdpb.proto
```diff
@@ -377,6 +378,9 @@ message BatchScanRegionsRequest {
 
     repeated KeyRange ranges = 3; // the given ranges must be in order.
     int32 limit = 4; // limit the total number of regions to scan.
+    // If output_must_contain_all_key_range is true, the output must contain all key ranges in the request.
+    // If the output does not contain all key ranges, the request is considered failed and returns an error(REGIONS_NOT_CONTAIN_ALL_KEY_RANGE).
+    bool output_must_contain_all_key_range = 5;
```
rleungx, line 383: Can we make it shorter?

okJiang: Do you have some better suggestions?😂

proto/pdpb.proto
```diff
@@ -377,6 +378,9 @@ message BatchScanRegionsRequest {
 
     repeated KeyRange ranges = 3; // the given ranges must be in order.
     int32 limit = 4; // limit the total number of regions to scan.
+    // If output_must_contain_all_key_range is true, the output must contain all key ranges in the request.
+    // If the output does not contain all key ranges, the request is considered failed and returns an error(REGIONS_NOT_CONTAIN_ALL_KEY_RANGE).
```
overvenus, line 382:
```suggestion
    // If output_must_contain_all_key_range is true, the output must contain all
    // key ranges in the request.
    // If the output does not contain all key ranges, the request is considered
    // failed and returns an error(REGIONS_NOT_CONTAIN_ALL_KEY_RANGE).
```

Prefer 80 character columns.

proto/pdpb.proto
rleungx: contain_all_key_range?

rleungx (approved):

proto/pdpb.proto
okJiang: What do you think which is better? @lance6716 @you06

you06: With the comment explaination, `contain_all_key_range` is clear enough to me.

okJiang: @you06 updated in https://github.com/pingcap/kvproto/pull/1252/commits/49bf52b9e60090df9f0eb3040a2ce7ae84edbe4f

okJiang:
如果没有其他问题的话，麻烦给我一个 approve

If you have no question, could you please give me an approve? @you06

you06 (approved):

ekexium:
We'd better merge such changes together with the related TiKV PRs to prevent breaking TiKV's build when using the master kvproto.

okJiang:
> We'd better merge such changes together with the related TiKV PRs to prevent breaking TiKV's build when using the master kvproto.

got it

PR #1253, opened 2024-07-22
persist region leader

rleungx:
The purpose of this PR is that we want to persist the leader information on the PD side so that if the PDs are reloaded, it won't affect the leader distribution in the cluster with a large number of regions. Because the heartbeat may be not processed in time and there will be redundant scheduling.
Besides, the region cache can get the leader ASAP.


proto/metapb.proto
```diff
@@ -127,6 +127,7 @@ message Region {
     bool is_in_flashback  = 7;
     // The start_ts that the current flashback progress is using.
     uint64 flashback_start_ts = 8;
+    Peer leader = 9;
```
overvenus, line 130: It's better to add a new protobuf message in pdpb.proto that wraps `metapb.Region` and `metapb.Peer` for better extensibility.
`metapb.Region` must only include states that need to be persisted in TiKV side, so "leader" should not be added here.

rleungx: It will cause breaking changes on the PD side. BTW, I think the purpose of metapb.Region are not the same on PD and TiKV.

Connor1996: +1, store `leader` in region meta is a little weird from the perspective of TiKV. Can PD store a separate key value for leader info?

rleungx: `leader` won't assign any value on the TiKV side.

Connor1996: I know, if there is no other better solution. I can accept reluctantly.

PR #1260, opened 2024-08-29
raft_serverpb: add the extra trait `use_plain_file` for SnapshotMeta

LykxSassinator:
## Description
This PR is a pre-requisite for TiKV #https://github.com/tikv/tikv/pull/17408/files, used for adding a new trait `use_plain_file` for SnapshotMeta. And this trait is introduced for marking that the  relative snapshot can be flushed with `plain file` or `sst files`.


pkg/configpb/configpb.pb.gw.go
```diff
@@ -31,7 +30,6 @@ var _ status.Status
 var _ = runtime.String
 var _ = utilities.NewDoubleArray
 var _ = descriptor.ForMessage
-var _ = metadata.Join
```
overvenus, line 34: Why is this file changed?

proto/raft_serverpb.proto
```diff
@@ -59,6 +59,9 @@ message SnapshotMeta {
     // It should only be used for v2 to send snapshot to v1.
     // See https://github.com/pingcap/tiflash/issues/7568
     uint64 commit_index_hint = 7;
+    // Flag reflects whether the snapshot is persisted with plain files
+    // or sst files. `true` means using plain files.
+    bool use_plain_file = 8;
```
overvenus, line 64: How about adding it to `SnapshotCFFile`? It's more flexible, eg. index regions have little data in the default CF but a large amount of data in the write CF.

pkg/configpb/configpb.pb.gw.go
LykxSassinator: It's cleaned by script automatically.

proto/raft_serverpb.proto
LykxSassinator: Reasonable. And the TODO -- "dump the plain file based on the size of each column family respectively" can be implemented based on it.

LykxSassinator:
To ensure the forward compatibility of TiKV, this change can be replaced with another implementation, by doing apply check on the `recv` side of TiKV.
So, this change is not necessary and can be safely closed.

PR #1261, opened 2024-09-02
runaway: Add processed_keys and request_ru to check 

HuSharp:
Part of https://github.com/pingcap/tidb/issues/54434.


nolouch (approved):

JmPotato (approved):

nolouch (approved):

JmPotato (approved):

proto/resource_manager.proto
```diff
@@ -211,6 +211,8 @@ enum RunawayWatchType {
 
 message RunawayRule {
   uint64 exec_elapsed_time_ms = 1;
+  int64 processed_keys = 2;
```
glorv, line 214: So only one of these fields can be valid at each time right? Maybe better add a `Type` to determine which one to use?

HuSharp: - We support subsets of all three elements. When all are valid, they are judged in order of time, keys, and ru. https://github.com/pingcap/tidb/pull/55147/files#diff-ad74786089837b88b4b3c3be8e011b4b30b73abe522accca049995f318f1f447R374
- I have added a rule field to show which threshold is in effect https://github.com/pingcap/tidb/pull/55147/files#diff-2fde3341e0231e7b162880b36f07d23b7949ca512c5536b50706ac5aa9072b25R630

glorv (approved):

PR #1264, opened 2024-09-04
resource_manager: add background resource limit

glorv:


nolouch (approved):

HuSharp:
can you paste the TiKV/TiDB related PR?

glorv:
> can you paste the TiKV/TiDB related PR?

Not submitted yet.

glorv:
@overvenus PTAL

proto/resource_manager.proto
```diff
@@ -231,7 +231,10 @@ message RunawaySettings {
 }
 
 message BackgroundSettings {
-  repeated string job_types = 1; 
+  // background task types.
+  repeated string job_types = 1;
+  // the percentage limit of total resource(cpu/io) that background tasks can use.
+  uint64 utilization_limit = 2;
```
overvenus, line 237: Should we use separated limits for IO and CPU?

glorv: Currently not, if both cpu and io are limited, the lower will take effect. BTW, in most cases, IO is not limited because it depends on the total IO quota which is not set by default.

HuSharp (approved):

glorv:
@Connor1996  @overvenus  PTAL

overvenus (approved):

PR #1269, opened 2024-09-17
br: add access key/secrets fields for aws kms

Tristan1900:
turns out we need to pass access key and secret access key as well for KMS. 
Currently TiKV loads them from environment which is not very flexible if more than one kms with different access keys specified in future


BornChanger (approved):

Tristan1900:
@overvenus could you take a quick look when you get a chance, sorry this is a last minute change before the new release so kind of in a hurry. Thanks!

overvenus (approved):

PR #1280, opened 2024-11-19
added versioning for migration

YuJuncen:
This PR added the vesrion and creator field to the `Migration` type for troubleshoting.


BornChanger (approved):

proto/brpb.proto
```diff
@@ -876,7 +876,17 @@ message DeleteSpansOfFile {
     uint64 whole_file_length = 3;
 }
 
+enum MigrationVersion {
+    // The initial version. Somehow "UNKNOWN".
+    M0 = 0;
```
Tristan1900, line 881: nit, maybe V0 and V1? usually use V as version,

proto/brpb.proto
```diff
@@ -876,7 +876,17 @@ message DeleteSpansOfFile {
...
+    // Added `creator` and `version`.
+    M1 = 1;
+}
+
 message Migration {
+    string creator = 5;
```
Tristan1900, line 887: nit, maybe sort in the numeric order, put them to the end of the message.

Tristan1900 (approved):
LGTM!

proto/brpb.proto
YuJuncen: protocol buffer requires the enum name is unique among *all* enums in the same package. The variant `V1` was used by `MetaFileVersion`...

proto/brpb.proto
YuJuncen: I think the `version` and `creator` field should be more primitive and essential than other fields. So I put them in the first line and people may notice them easier.

3pointer (approved):

PR #1281, opened 2024-12-02
Add ext full backup

YuJuncen:


proto/brpb.proto
```diff
@@ -898,6 +898,48 @@ message MetaEdit {
...
+// RewrittenTableID records a sort of modification over the SSTs during restoring.
+// 
+// When "backing up" this "restored" SST, we want to backup what it was "restored".
+// But in some cases, SSTs are not restored "as is", TiKV may rewrite or truncate its
+// content by the client's request. 
+message RewrittenTableID {
+    // The table ID in the backup data.
+    int64 upstream_of_upstream = 1;
+    // The rewritten table ID during restoring.
+    int64 upstream = 2;
```
BornChanger, line 910:
```suggestion
    int64 table_id_in_ancestor_upstream = 1;
    // The rewritten table ID during restoring.
    int64 table_id_in_upstream_table_id = 2;
```

proto/brpb.proto
```diff
@@ -898,6 +898,48 @@ message MetaEdit {
     bool destruct_self = 4;
 }
 
+// RewrittenTableID records a sort of modification over the SSTs during restoring.
+// 
+// When "backing up" this "restored" SST, we want to backup what it was "restored".
```
Tristan1900, line 903: I feel like `we want to back up the original SST before any rewrite` maybe more intuitive?

proto/brpb.proto
Tristan1900: agree, but we may need to explain what `ancestor` is, as it's a new term.

proto/brpb.proto
```diff
@@ -898,6 +898,48 @@ message MetaEdit {
...
+    // The hint of the common prefix of the files.
+    // Used for speed up garbage collecting.
+    string files_prefix_hint = 4;
+    // Whether this piece of extra full backup is finished.
+    // If not, `as_if_ts` should be ignored and this should only
+    // be restored when there is a finished backup with the same
+    // `backup_uuid`.
+    bool finished = 5;
+    // When checkpoint enabled, one restoration may be
+    // sperated to many of `ExtraFullBackup`s.
```
Tristan1900, line 935: a typo `sperated `

proto/brpb.proto
```diff
@@ -898,6 +898,48 @@ message MetaEdit {
...
+message RewrittenTableID {
+    // The table ID in the backup data.
+    int64 upstream_of_upstream = 1;
+    // The rewritten table ID during restoring.
+    int64 upstream = 2;
+}
+
+// ExtraFullBackup is created by BR restoring to a cluster with PITR enabled.
+// This is uploaded by the BR CLI.
+message ExtraFullBackup {
```
Tristan1900, line 915: What do you think of the name `LogBackupInjestedSST` or other similar options. 
I feel like in our code base there are a lot vague terms and we need to remember what they mean when reading the code, making a bit hard to understand. I personally like to name as explicit as possible. Let me know your thoughts. 
Feel free to ignore the comment if you think otherwise!

Tristan1900 (approved):
LGTM!

proto/brpb.proto
YuJuncen: It seems there are too many repeated `table_id`s. I think `ancestor_upstream` and `upstream` would be fine.

BornChanger (approved):

PR #1282, opened 2024-12-05
br: add more fields for merged backup SST

Leavrth:


proto/brpb.proto
```diff
@@ -147,6 +147,14 @@ message BackupRange {
     repeated File files = 3;
 }
 
+// The message saves the physical table checksum
+message TableMeta {
+    int64 physical_id = 1;
```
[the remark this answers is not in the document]
Leavrth: physical id = partition id or table id

proto/brpb.proto
```diff
@@ -147,6 +147,15 @@ message BackupRange {
     repeated File files = 3;
 }
 
+// The message saves the physical table checksum
+message TableMeta {
```
Tristan1900, line 151: feels like the name could be TableChecksum unless if you want to add more in the future?

Leavrth: Actually I named it `TableChecksum` at first, but now I prefer to named it `TableMeta`. That's because `TableMeta`  is also used to mark the real kvs in the SST file.
1. If we have backed up an SST file like this:
[   table0:[kv1, kv2, ...], table1:[kv1, kv2, ...], ...   ]
However, another SST file is backed up and is overlapped with the above SST file.
[ table1:[kv1, kv2, ...], ...   ]
Therefore, BR remove the TableMeta::table1 in the first SST file metafile and now `TableMeta` shows which kvs in the SST file is actually valid.
2. When BR filters out some tables, BR also removes these `TableMeta`s to mark the skipped kvs.

BornChanger (approved):

BornChanger (approved):

YuJuncen (approved):

YuJuncen (approved):

PR #1298, opened 2025-03-24
add failed_regions_id for scatter region response

GMHDBJD:
ref https://github.com/tikv/pd/issues/9088
add failed_regions_id for scatter region response, so client can retry for the failed regions.


okJiang (approved):

proto/pdpb.proto
```diff
@@ -1228,6 +1229,8 @@ message SplitAndScatterRegionsResponse {
     uint64 split_finished_percentage = 2;
     uint64 scatter_finished_percentage = 3;
     repeated uint64 regions_id = 4;
+    // scatter failed region ids
+    repeated uint64 failed_regions_id = 5;
```
rleungx, line 1233: Do we need to distinguish between failing to scatter and failing to split?

rleungx (approved):

PR #1301, opened 2025-04-17
add edition proto for dedicated build

zyguan:
There are some compatibility issues between master and dedicated. To reduce maintenance costs, this PR try to introduce conditional compilation instead of maintaining [the standalone branch](https://github.com/pingcap/tidb/tree/dedicated).


proto/edition.proto
```diff
@@ -0,0 +1,20 @@
...
+option (gogoproto.goproto_unrecognized_all) = false;
+option (gogoproto.goproto_sizecache_all) = false;
+option (rustproto.lite_runtime_all) = true;
+
+option java_package = "org.tikv.kvproto";
+
+enum EditionType {
+    STANDARD = 0;
+    DEDICATED = 1;
+}
```
D3Hunter, line 20: those are not good term to describe for first/next generation

zyguan:
the issue is resolved by https://github.com/pingcap/tidb/pull/60668

PR #1308, opened 2025-04-23
kvrpcpb: Add more fields in CommitRequest/Response to give more lock info

lcwangchao:
related to https://github.com/tikv/client-go/issues/1631

1. Add field `primary_key` to `CommitRequest` so that the TiKV side can determine whether the current request is committing the primary. We can do some different logging when it is primary commit or not.
2. Add field `TxnLockNotFound` to `KeyErr` so that the client can get the actual missing lock key when lock is not found


proto/kvrpcpb.proto
```diff
@@ -386,6 +386,11 @@ message CommitRequest {
     repeated bytes keys = 3;
     // Timestamp for the end of the transaction. Must be greater than `start_version`.
     uint64 commit_version = 4;
+    // primary_key indicates the primary key of the transaction.
+    // If it is empty, it means the primary key is not set and we cannot determine the whether it is the primary commit.
+    // Otherwise if primary_key is present, we can determine the commit is primary or not by checking whether the
+    // primary_key is in the keys.
+    bytes primary_key = 6;
```
lcwangchao, line 393: Another option is to add a field `is_primary_commit`. But I think it may introduce some compatibility issues because `primary_key` is always `false` for clients with old versions.

proto/kvrpcpb.proto
```diff
@@ -1084,6 +1090,10 @@ message PrimaryMismatch {
     LockInfo lock_info = 1;
 }
 
+message TxnLockNotFound {
+    bytes key = 1;
+}
+
```
lcwangchao, line 1096: Do we need to add an extra message `DebugInfo` which contains some debugging info such as mvcc info here?

cfzjywxk: Perhaps it is ok to return a single key mvcc info in the response, @MyonKeminta @ekexium @zyguan What do you think?

lcwangchao: In this, I does not include mvcc info in TxnLockNotFound because it seems mvcc info is not necessary for all scenes. The client should not have an assumption that it is present all the time.

zyguan (approved):
The changes LGTM, however I'm not sure how it could be used as tidb would already print logs like `Error(Txn(Error(Mvcc(Error(TxnLockNotFound { start_ts: TimeStamp(456922239593087375), commit_ts: TimeStamp(456922240510592670), key: [116, 128, 0, 0, 0, 0, 1, 203, 241, 95, 114, 184, 0, 0, 14, 97, 242, 19, 198] })))))` .

lcwangchao:
> The changes LGTM, however I'm not sure how it could be used as tidb would already print logs like `Error(Txn(Error(Mvcc(Error(TxnLockNotFound { start_ts: TimeStamp(456922239593087375), commit_ts: TimeStamp(456922240510592670), key: [116, 128, 0, 0, 0, 0, 1, 203, 241, 95, 114, 184, 0, 0, 14, 97, 242, 19, 198] })))))` .

When the above error returns to TiDB, the client can only get the string bytes of the above message. Though we can parse it to get the conflicted key, it is tricky because the "message" format is not guaranteed not to change in the future.

proto/kvrpcpb.proto
```diff
@@ -1084,6 +1090,11 @@ message PrimaryMismatch {
     LockInfo lock_info = 1;
 }
 
+message TxnLockNotFound {
+    bytes key = 1;
+    MvccInfo mvcc_info = 2;
```
lcwangchao, line 1095: I added an extra field `mvcc_info` here for client debugging

lcwangchao: Another option is to add new message:

```
message DebugInfo {
    MvccInfo mvcc_info = 1;
}
```

ekexium (approved):

proto/kvrpcpb.proto
```diff
@@ -1020,6 +1036,10 @@ message KeyError {
     CommitTsTooLarge commit_ts_too_large = 9; // Calculated commit TS exceeds the limit given by the user.
     AssertionFailed assertion_failed = 10; // Assertion of a `Mutation` is evaluated as a failure.
     PrimaryMismatch primary_mismatch = 11; // CheckTxnStatus is sent to a lock that's not the primary.
+    TxnLockNotFound txn_lock_not_found = 12; // TxnLockNotFound indicates the txn lock is not found.
+
+    // Extra information for error debugging
+    DebugInfo debug_info = 100;
```
ekexium, line 1042: Why let it be 100?

lcwangchao: to separate `DebugInfo` with certain error structs. `DebugInfo` is a common field that multiple error type may need it.

proto/kvrpcpb.proto
```diff
@@ -386,6 +396,12 @@ message CommitRequest {
     repeated bytes keys = 3;
     // Timestamp for the end of the transaction. Must be greater than `start_version`.
     uint64 commit_version = 4;
+    // commit_role indicates the current commit request is a primary commit or a secondary commit.
+    // It's value maybe `Unknown` when using a client with an old version.
+    CommitRole commit_role = 6;
+    // primary_key indicates the primary key of the transaction.
+    // It's value maybe empty when using a client with an old version.
```
cfzjywxk, line 403:
```suggestion
    // Its value may be empty when using an old version client.
```

cfzjywxk (approved):

ekexium (approved):

cfzjywxk (approved):

PR #1310, opened 2025-05-13
kvrpcpb: add `UndeterminedResult` to `KeyError`

lcwangchao:
If the client received the `UndeterminedResult` error, the current transaction's commit state is not determined, and the TiDB side should disconnect the current connection.


MyonKeminta (approved):

cfzjywxk:
@MyonKeminta PTAL again

cfzjywxk (approved):

PR #1311, opened 2025-05-16
feat(resource_manager): add keyspace support

JmPotato:
Ref https://github.com/tikv/pd/issues/9296.


JmPotato:
@guo-shaoge @lhy1024 @glorv PTAL

proto/resource_manager.proto
```diff
@@ -28,6 +28,7 @@ service ResourceManager {
 
 message ListResourceGroupsRequest{
   bool with_ru_stats = 1;
+  uint32 keyspace_id = 2;
```
glorv, line 31: For tikv/tiflash, they need to get all resource groups of all keyspaces, how to express this?

proto/resource_manager.proto
```diff
@@ -38,6 +39,7 @@ message ListResourceGroupsResponse{
 message GetResourceGroupRequest {
   string resource_group_name = 1;
   bool with_ru_stats = 2;
+  uint32 keyspace_id = 3;
```
glorv, line 42: For a cluster that does not enable keyspace, the keyspace id will be u32::MAX after this change, but 0 with the old proto. So in the rolling upgrade scenario, how to distinct between them? Note that 0 is also a valid keyspace id.

proto/resource_manager.proto
lhy1024, line 31: Do we need to provide keyspace name?

JmPotato: Keyspace ID is its unique identifier; it seems unnecessary to introduce a name.

glorv:
@lhy1024 @nolouch PTAL

nolouch (approved):

PR #1314, opened 2025-05-21
feat(resource_manager): remove ListKeyspaceResourceGroups RPC

JmPotato:
Ref https://github.com/tikv/pd/issues/9296.

No need to introduce `ListKeyspaceResourceGroups` because:

- The old version without setting `KeyspaceID`, returning all resource groups, but since it is an old version, it should only include legacy groups under `NullKeyspace`, so returning all is the same as the previous behavior.
- The new version without setting `KeyspaceID`, returning all resource groups, including `NullKeyspace` and all other keyspaces. Because it is a new version, TiDB will aggregate based on the keyspace, avoiding issues with duplicate group names.
- The new version with setting `KeyspaceID`, returning all resource groups corresponding to the specified keyspace ID, representing the new behavior.


lhy1024 (approved):

lhy1024 (approved):

glorv (approved):

glorv:
@nolouch PTAL

nolouch (approved):

rleungx (approved):

lhy1024 (approved):

glorv (approved):

PR #1320, opened 2025-06-04
Add keyspace_name in kvrpc context

yibin87:
https://github.com/pingcap/kvproto/issues/1322


proto/kvrpcpb.proto
```diff
@@ -910,6 +910,10 @@ message Context {
     // Some information used for resource control.
     ResourceControlContext resource_control_context = 28;
 
+    // The keyspace that the request is sent to.
+    // NOTE: This field is only meaningful while the api_version is V2.
+    string keyspace_name = 31;
```
Connor1996, line 915: Why is there no 29-31 field number before?

yibin87: I don't know the reason, from history, when keyspace_id was first added, 29-31 is skipped: https://github.com/pingcap/kvproto/commit/c6df78cc9dea61bdf913ec8bac49e245980c1a94
And these numbers are not reserved.

cfzjywxk:
@yibin87 Please add the issue reference link in the description

yibin87:
> @yibin87 Please add the issue reference link in the description

Done.

cfzjywxk (approved):

cfzjywxk:
@yibin87 @Connor1996 
Is the `keyspace_name` ensured to be globally unique? If so better to add comments about it.

And why not using the `keyspace_id` instead?

Not quite clear about the designs of `keyspace_id` and `keyspace_name`.

yibin87:
> @yibin87 @Connor1996 Is the `keyspace_name` ensured to be globally unique? If so better to add comments about it.
> 
> And why not using the `keyspace_id` instead?
> 
> Not quite clear about the designs of `keyspace_id` and `keyspace_name`.

No, keyspace_name is not expected to be globally unique. Keyspace_id maybe globally unique. However, we design to expose keyspace_name from kernerl instead of keyspace_id, and it is not easy to get keyspace_name from keyspace_id inside tikv, thus we pass it in grpc request here.

PR #1325, opened 2025-06-12
resource_control: add cross az traffic bytes in ru consumption

glorv:


glorv:
@JmPotato @rleungx  PTAL

proto/resource_manager.proto
```diff
@@ -171,6 +171,8 @@ message Consumption {
   double sql_layer_cpu_time_ms = 6;
   double kv_read_rpc_count = 7;
   double kv_write_rpc_count = 8;
+  uint64 read_cross_az_traffic_bytes = 9;
```
rleungx, line 174: Do we need to use double?

nolouch (approved):

proto/resource_manager.proto
glorv: No, actually all these counters (except the rru and wru) should be defined with uint64 as they are always integer, but replacing double with int is incompatible change, so I replace with uint64 only for new fields.

rleungx (approved):

PR #1338, opened 2025-08-14
metaedit: add new field to record meta files that can be skipped

3pointer:
Description:

This PR introduces a new field all_data_files_handled to the MetaEdit protobuf definition. The purpose of this field is to explicitly record whether all data KV files referenced by the meta have already been handled, allowing the compaction to safely skip unnecessary processing in subsequent meta files. To save the cost of reading s3.


proto/brpb.proto
```diff
@@ -900,6 +900,9 @@ message MetaEdit {
     repeated DeleteSpansOfFile delete_logical_files = 3;
     // Whether to delete the metadata itself.
     bool destruct_self = 4;
+
+    // Whether all data kv files have been handled (applied or ignored).
+    bool all_data_files_handled = 5;
```
YuJuncen, line 905: `all_data_files_deleted`?

YuJuncen (approved):

PR #1345, opened 2025-09-04
graceful shutdown implement (#17221)

hujiatao0:
Adding an is_stopping status flag to the StoreHeartbeat message to support the graceful shutdown of TiKV


proto/pdpb.proto
```diff
@@ -776,9 +776,11 @@ message StoreStats {
     uint64 total_memory = 28;
     // Used memory of the store in bytes.
     uint64 used_memory = 29;
+    // If the store is stopping.
+    bool is_stopping = 30;
     // Network_slow_scores indicate the network status between TiKV nodes, ranging from 1 to 100 (lower is better).
     // StoreID -> score
-    map<uint64, uint64> network_slow_scores = 30;
+    map<uint64, uint64> network_slow_scores = 31;
```
overvenus, line 783: Please do not change existing tag number, it breaks backward compatibility. 
```suggestion
    map<uint64, uint64> network_slow_scores = 30;
```

proto/pdpb.proto
```diff
@@ -781,6 +781,9 @@ message StoreStats {
     map<uint64, uint64> network_slow_scores = 30;
     // The statistics about DFS uploads.
     repeated DfsStatItem dfs = 31;
+    // If the store is stopping.
```
hbisheng, line 784:
```suggestion
    // True if the store is undergoing graceful shutdown.
```

PR #1351, opened 2025-09-16
pd: support disable-auto-split region

HunDunDM:


lhy1024 (approved):

lhy1024:
ref https://github.com/tikv/pd/issues/9764

proto/pdpb.proto
```diff
@@ -643,6 +647,7 @@ message RegionHeartbeatResponse {
     //       2. demoting voter directly
     ChangePeerV2 change_peer_v2 = 9;
     BatchSwitchWitness switch_witnesses = 10;
+    ChangeAutoSplit change_auto_split = 11;
```
rleungx: Will ChangeAutoSplit have more fields?

proto/pdpb.proto
```diff
@@ -673,11 +678,18 @@ message ReportSplitResponse {
     ResponseHeader header = 1;
 }
 
+enum AutoSplitReason {
```
rleungx: How about
```suggestion
enum SplitReason {
```

PR #1370, opened 2025-11-06
add field trace_control_flags for tracing

ekexium:
ref pingcap/tidb#64008


ekexium:
The PR needs further consideration. hold on

proto/kvrpcpb.proto
```diff
@@ -936,6 +936,10 @@ message Context {
 
     // The trace id of the request, will be used for tracing the request's execution's inner steps.
     bytes trace_id = 36;
+
+    // If true, TiKV should log this trace immediately without buffering.
+    // This field is set by client-go based on an extractor function provided by TiDB.
+    bool force_immediate_log = 37;
```
cfzjywxk: Is it ok to re-use the `trace_id` field above?

PR #1379, opened 2025-11-21
pd: add callee_id to RequestHeader

iosmanthus:
`callee_id` is used to indicate the expected server ID to receive the request. The server **SHOULD** check if the `callee_id` matches its own ID, normally it's the client url address of the server. We encountered an issue where a tso client caches the stale DNS result and the new tso server takes over the old server's network address. Then, the tso client connects to the new tso server (a follower) with the actual leader's hostname, creating a hallucination that the leader is not the leader anymore. If the client meets this case, it might never get out of it if it doesn't refresh its DNS cache by reconnecting. ref https://github.com/tikv/pd/issues/9894


disksing (approved):

bufferflies (approved):

ystaticy (approved):

bufferflies:
@copilot resolve the merge conflicts in this pull request

iosmanthus:
@bufferflies @ystaticy PTAL

bufferflies:
@rleungx PTAL

rleungx (approved):

PR #1450, opened 2026-04-13
move metaversion from tidb to kvproto

JoyC-dev:


Leavrth (approved):

proto/brpb.proto
```diff
@@ -143,13 +144,17 @@ message BackupMeta {
 
   // record the size of the backup data files and meta files
   uint64 backup_size = 25;
+  // NOTE: if adding new fields here affects backup meta format/semantics,
```
BornChanger: Is the comment automatically added?

JoyC-dev: proto/brpb.proto is not the auto-generated file (pkg/*/*.pb.go does) , so I manually added it. Besides, I added notes to the place can possibly add metadata.

BornChanger (approved):

PR #1457, opened 2026-04-27
coprocessor: add TiCI estimate count RPC

wshwsh12:
## Summary
- Add TiCI estimate count request/response messages to coprocessor proto.
- Add the GetEstimateTiCICount service RPC after GetTiFlashSystemTable.
- Regenerate Go protobuf bindings and proto lock metadata.


proto/coprocessor.proto
```diff
@@ -108,6 +108,22 @@ message TableShardInfos {
...
 }
 
+message TiCIEstimateCountRequest {
+  kvrpcpb.Context context = 1;
+  uint64 start_ts = 2;
+  int64 table_id = 3;
+  int64 index_id = 4;
+  bytes fts_query_info = 5;
+  string time_zone_name = 6;
+  int64 time_zone_offset = 7;
```
wshwsh12, line 118: These two fields are needed because this estimate RPC carries the TiCI/FTS expression outside the normal DAG request path, so TiFlash would otherwise not know the TiDB session timezone when reconstructing the evaluation context. `time_zone_name` preserves named zones such as `Asia/Shanghai` (including DST rules when applicable), while `time_zone_offset` is the fallback for fixed-offset/session-local zones. Keeping both matches the existing DAG request timezone semantics and prevents timezone-sensitive expression evaluation from silently using the TiFlash process default timezone.

proto/tikvpb.proto
```diff
@@ -126,6 +126,8 @@ service Tikv {
 
   /// Get system table from TiFlash
   rpc GetTiFlashSystemTable(kvrpcpb.TiFlashSystemTableRequest) returns (kvrpcpb.TiFlashSystemTableResponse) {}
+  /// Get estimate tici count from tiflash
```
cfzjywxk:
```suggestion
  /// Get estimate tici count from TiFlash
```
Better to keep the naming convention

cfzjywxk (approved):

windtalker (approved):
lgtm

PR #1458, opened 2026-04-30
add API V3 keyspace identity protos

disksing:
Issue: #1506

Summary
- Add public API V3 KeyspaceIdentity schema and wire it through KV, PD, TSO, keyspace, GC, and related tenant-scoped protobuf APIs.
- Preserve legacy keyspace_id fields for V1/V2 compatibility while documenting V3 identity semantics.
- Update generated bindings and proto metadata for downstream API V3 consumers.

Validation
- make proto-fmt
- go build ./pkg/...
- go test ./pkg/...
- make check reached the expected proto.lock change prompt
- cargo check is still blocked locally by the known libz/Xcode header issue


proto/pdpb.proto
```diff
@@ -816,9 +823,14 @@ message DfsStatScope {
...
   bool is_global = 1;
   // The keyspace of this statistic. Ignore when is_global is true.
+  // NOTE: This field is only meaningful for V1/V2 compatibility. V3 should use keyspace_identity.
   uint32 keyspace_id = 2;
   // The component that provides the statistic.
   string component = 3;
+  // V3 keyspace identity of this statistic. Ignore when is_global is true or keyspace_identities is set.
+  apipb.KeyspaceIdentity keyspace_identity = 4;
+  // V3 multi-keyspace statistic scope. Ignore when is_global is true.
+  repeated apipb.KeyspaceIdentity keyspace_identities = 5;
```
sunxiaoguang:
```suggestion
  repeated apipb.KeyspaceIdentity keyspace_identity = 4;
```

proto/autoid.proto
```diff
@@ -20,7 +21,10 @@ message AutoIDRequest {
   uint64 n = 4;
   int64 increment = 5;
   int64 offset = 6;
+  // V1/V2 compatibility keyspace id. V3 should use keyspace_identity.
```
sunxiaoguang:
```suggestion
  // V1/V2 compatibility keyspace id. V3 should use keyspace_identity.
  oneof keyspace {
    uint32 keyspaceID = 7;
    apipb.KeyspaceIdentity identity = 8;
  }
```

proto/keyspacepb.proto
```diff
@@ -43,13 +49,35 @@ enum KeyspaceState {
...
+  // If unset, V3 name-only lookup should use LookupKeyspace and may return multiple keyspaces.
+  uint32 namespace_id = 3;
 }
 
 message LoadKeyspaceResponse {
   pdpb.ResponseHeader header = 1;
   KeyspaceMeta keyspace = 2;
 }
 
+message LookupKeyspaceRequest {
```
sunxiaoguang: Can we find out some scenarios when looking up keyspaces with a given name across multiple namespaces make senses.

proto/pdpb.proto
```diff
@@ -1005,13 +1017,18 @@ message UpdateServiceGCSafePointResponse {
...
   uint32 keyspace_id = 2;
+  // V3 keyspace identity.
+  apipb.KeyspaceIdentity keyspace_identity = 3;
 }
 
 message GetGCSafePointV2Response {
   ResponseHeader header = 1;
 
   uint64 safe_point = 2;
+  // V3 keyspace identity served by this response.
```
sunxiaoguang: Is the keyspace identity helpful when it is already known when sending the request

proto/pdpb.proto
```diff
@@ -1035,19 +1055,25 @@ message WatchGCSafePointV2Response {
...
+  // V3 keyspace identity.
+  apipb.KeyspaceIdentity keyspace_identity = 4;
 }
 
 message UpdateGCSafePointV2Response {
   ResponseHeader header = 1;
 
   uint64 new_safe_point = 2;
+  // V3 keyspace identity served by this response.
+  apipb.KeyspaceIdentity keyspace_identity = 3;
```
sunxiaoguang: Is the keyspace identity helpful when it is already known when sending the request

proto/pdpb.proto
```diff
@@ -1065,15 +1093,20 @@ message UpdateServiceSafePointV2Response {
   bytes service_id = 2;
   int64 ttl = 3;
   uint64 min_safe_point = 4;
+  // V3 keyspace identity served by this response.
+  apipb.KeyspaceIdentity keyspace_identity = 5;
```
sunxiaoguang: Is the keyspace identity helpful when it is already known when sending the request

proto/tsopb.proto
```diff
@@ -40,17 +44,22 @@ message RequestHeader {
...
   uint64 cluster_id = 1;
   Error error = 2;
 
-  // keyspace_id is the unique id of the tenant/keyspace as the response receiver.
+  // keyspace_id is the unique id of the tenant/keyspace as the response receiver in V1/V2.
+  // V3 should use keyspace_identity and must not read this legacy field as the full identity.
   uint32 keyspace_id = 3;
   // keyspace_group_id is the unique id of the keyspace group to which the tenant/keyspace belongs.
   uint32 keyspace_group_id = 4;
+  // V3 keyspace identity served by this response.
```
sunxiaoguang: Is the keyspace identity helpful when it is already known when sending the request

proto/tsopb.proto
```diff
@@ -18,6 +19,8 @@ service TSO {
   rpc Tso(stream TsoRequest) returns (stream TsoResponse) {}
   // Find the keyspace group that the keyspace belongs to by keyspace id.
   rpc FindGroupByKeyspaceID(FindGroupByKeyspaceIDRequest) returns (FindGroupByKeyspaceIDResponse) {}
+  // Find the keyspace group that the keyspace belongs to by V3 keyspace identity.
```
sunxiaoguang: How about extend FindGroupByKeyspaceIDRequest to include KeyspaceIdentity.

proto/keyspacepb.proto
```diff
@@ -72,8 +103,13 @@ message UpdateKeyspaceStateResponse {
 
 message GetAllKeyspacesRequest {
   pdpb.RequestHeader header = 1;
+  // V1/V2 compatibility pagination cursor. V3 should use namespace_id and start_keyspace_identity.
   uint32 start_id = 2;
   uint32 limit = 3;
+  // V3 namespace-limited pagination. Must be non-zero in V3.
```
sunxiaoguang: namespace-scoped pagnination

proto/kvrpcpb.proto
```diff
@@ -939,9 +940,16 @@ message Context {
   // NOTE: This field is only meaningful while the api_version is V2.
   string keyspace_name = 31;
 
-  // The keyspace that the request is sent to.
-  // NOTE: This field is only meaningful while the api_version is V2.
-  uint32 keyspace_id = 32;
+  oneof keyspace {
```
rleungx, line 943: Changing this existing scalar field into a `oneof` is source-incompatible for generated Go users even though the wire tag is preserved. `Context.KeyspaceId` disappears, and downstream PD/TiDB/client-go code constructs or reads this field directly, so upgrading kvproto would break V1/V2 callers that do not use V3. The legacy scalar should stay as a normal field and `keyspace_identity` should be added alongside it, as was done for `AutoIDRequest` and `KeyspaceMeta`.

disksing: Thanks, fair point. I originally kept `keyspace_id` as a normal field and added `keyspace_identity` alongside it to minimize downstream Go source changes. Later I changed it to `oneof` based on review feedback, since semantically it is clearer that only one keyspace identifier should be set.

I think both directions are reasonable: parallel fields are more source-compatible, while `oneof` is clearer semantically. @sunxiaoguang what do you think? If we prefer minimizing downstream changes for `kvrpcpb.Context`, I can change this back to the parallel-field style.

sunxiaoguang: Thanks @rleungx @disksing for the insightful discussion.
I lean towards the `oneof` approach here. Here is my perspective:
Confined Blast Radius: While changing to oneof is indeed source-incompatible, the generated protobuf code is primarily encapsulated within our rich clients (**client-go, client-rust, client-c**). The blast radius of this **incompatible change is limited and well-contained within these repositories**, rather than leaking indefinitely downstream.
Compiler-Guaranteed Safety: For strongly-typed languages, breaking the source code during compilation is actually a major benefit in this context. It forces the compiler to **help us pinpoint every single callsite** that accesses the relevant fields. This **completely eliminates the risk of missing any field migrations**, which could **otherwise lead to silent, hard-to-debug runtime issues**.
Semantics vs. Refactoring Cost: Maintaining parallel fields compromises semantic clarity and risks inconsistent states (e.g., both fields being set). The value of absolute correctness and explicit semantics brought by oneof far outweighs the cost of limited code adjustments in the client libraries.
Therefore, I'd prefer to go with oneof to ensure long-term robustness.

proto/kvrpcpb.proto
```diff
@@ -1037,6 +1045,12 @@ enum APIVersion {
   // V2 server accpets V2 requests and V1 transactional requests that statrts with TiDB key
   // prefix (`m` and `t`).
   V2 = 2;
+  //
+  // `V3` uses user-key wire semantics for normal KV RPCs. Servers encode an
+  // 8-byte physical prefix at the serving boundary:
+  // mode(1) + namespace_id(4 bytes, big endian) + keyspace_id(3 bytes, big endian).
```
overvenus, line 1051: What's the semantic difference between namespace id and keyspace id?
What are the considerations of encoding namespace id first? Can it be `mode(1) + keyspace_id(3 bytes, big endian) + namespace_id(4 bytes, big endian)`?

coocood:
@disksing 
Do we have dev branch of tidb & tikv that compiles with API V3?

proto/kvrpcpb.proto
disksing: namespace_id represents the higher-level namespace that groups multiple keyspaces. We intentionally encode it before keyspace_id because storage keys are ordered lexicographically. Using mode + namespace_id + keyspace_id ensures that all keyspaces belonging to the same namespace form one contiguous key range after sorting. If keyspace_id were encoded first, keyspaces from the same namespace would be scattered across different ranges. Therefore, we prefer to keep the namespace ID before the keyspace ID.

disksing:
> @disksing Do we have dev branch of tidb & tikv that compiles with API V3?

PRs update kvproto and fix compile problem.

TiKV：[tikv/tikv#19857](https://github.com/tikv/tikv/pull/19857)
PD：[tikv/pd#11024](https://github.com/tikv/pd/pull/11024)
client-go：[tikv/client-go#2024](https://github.com/tikv/client-go/pull/2024)
TiDB：[pingcap/tidb#69942](https://github.com/pingcap/tidb/pull/69942)

coocood (approved):

proto/keyspacepb.proto
```diff
@@ -41,9 +52,72 @@ enum KeyspaceState {
...
+message UpdateNamespaceStateResponse {
+  pdpb.ResponseHeader header = 1;
+  NamespaceMeta namespace = 2;
+}
+
 message LoadKeyspaceRequest {
   pdpb.RequestHeader header = 1;
   string name = 2;
+  // V3 namespace-scoped name lookup. If unset, V3 name-only lookup should use LookupKeyspace and may return multiple keyspaces.
+  NamespaceRef namespace = 3;
```
rleungx, line 120: `LoadKeyspaceByIDRequest` cannot identify a V3 keyspace because it carries no namespace.

disksing: fixed

proto/keyspacepb.proto
```diff
@@ -26,7 +32,12 @@ service Keyspace {
 }
 
 message KeyspaceMeta {
-  uint32 id = 1;
+  oneof keyspace {
+    // V1/V2 compatibility keyspace id. V3 should read keyspace_identity instead.
+    uint32 id = 1;
+    // Canonical V3 keyspace identity.
+    apipb.KeyspaceIdentity keyspace_identity = 8;
```
rleungx, line 39: `externalworkloadpb.RequestHeader` still carries only `keyspace_id` and `keyspace_name`, and TiDB constructs it from `KeyspaceMeta.GetId()`/`GetName()`. A V3 meta selecting this arm makes `GetId()` return 0, so same-named keyspaces in different namespaces can become indistinguishable and GC/TTL/background workload events can be routed or recycled for the wrong tenant.

proto/autoid.proto
```diff
@@ -20,7 +21,12 @@ message AutoIDRequest {
   uint64 n = 4;
   int64 increment = 5;
   int64 offset = 6;
-  uint32 keyspaceID = 7;
+  oneof keyspace {
```
rleungx, line 24: `RebaseRequest` needs the same keyspace discriminator. A stale client connected to an address that restarted for another keyspace can otherwise rebase the wrong tenant. V3 should carry the full keyspace identity here as well.

proto/keyspacepb.proto
disksing: fixed

proto/autoid.proto
disksing: fixed

proto/resource_manager.proto
```diff
@@ -27,7 +27,7 @@ service ResourceManager {
 }
 
 // KeyspaceIDValue is a wrapper for the value of keyspace ID.
-// Because the 0 value is a valid keyspace ID, we need to use a wrapper to distinguish it from the null keyspace ID.
+// Because the 0 value is a valid keyspace ID in V1/V2, we need to use a wrapper to distinguish it from the null keyspace ID.
```
rleungx: `ResourceManager` still cannot distinguish V3 keyspaces that share the same numeric ID across namespaces. The V3 PD client sends only `identity.keyspace_id` through these messages, and the RM server indexes resource groups and token buckets by that value, so those tenants would share resource-group metadata and RU quota.

rleungx (commented):
The rest LGTM.

rleungx (approved):

windtalker (approved):
lgtm

cfzjywxk (approved):

bb7133 (approved):
LGTM

zhangjinpeng87 (approved):
LGTM

proto/pdpb.proto
sunxiaoguang, line 236:
```suggestion
// This message intentionally omits namespace/keyspace-related fields because API v2 never supported keyspaces when routing TSO requests through the PD API server.
message TsoRequest {
```

proto/keyspacepb.proto
```diff
@@ -16,7 +17,12 @@ option (rustproto.lite_runtime_all) = true;
 
 // Keyspace provides services to manage keyspaces.
 service Keyspace {
```
niubell, line 19: Namespace contains keyspace, but we put Namespace in the Keyspace service. Maybe adding a new Namespace service is better?

disksing: Namespace and keyspace are tightly coupled in API V3, and namespace_id is part of KeyspaceIdentity. Therefore, keeping their management APIs in the same service seems reasonable. We can revisit a separate Namespace service if it gains a more independent lifecycle later.

yudongusa:
Why didn't any reviewer point out the missing github related issue and design spec?

PR #1480, opened 2026-06-01
br, import: reserve a range of field number for downstream fork

lance6716:


lance6716:
/cc @BornChanger

We have implemented a feature for BR at downstream. To avoid RPC conflict, we want to reserve a range

YuJuncen (approved):

BornChanger (approved):

PR #1494, opened 2026-06-24
resource_manager: add signed RU settlement report

YuhaoZhang00:
## Summary
- Add TokenBucketRequest.settlement_since_last_request for signed RRU/WRU token settlement reporting.
- Clarify consumption_since_last_request as actual resource usage for stats and metering.

## Related
- Supports tikv/pd#10611.


YuhaoZhang00:
Closing this companion PR because the PD change is being minimized to avoid adding settlement_since_last_request; PD will keep using consumption_since_last_request with its existing non-decreasing semantics.

PR #1497, opened 2026-07-14
coprocessor: add batch task data merge negotiation fields

0xPoe:
### What problem does this PR solve?

TiKV can reduce batched analyze response traffic by merging per-task results into `Response.data`. However, an empty `batch_responses` cannot distinguish merged results from unprocessed tasks, which can cause double counting and lose execution details. Besides, batching more tasks into one request multiplies per-store concurrency, so clients also need a way to limit it.

### What is changed and how it works?

Add explicit negotiation:

- `Request.allow_batch_task_data_merge` lets the client opt in.
- `StoreBatchTaskResponse.data_merged_into_response` confirms that a task's data was merged while preserving its `exec_details_v2`.
- `Request.execute_batch_tasks_serially` lets the client ask the store to run the primary task and batched tasks one at a time (execution order is not guaranteed).

Without client opt-in or store support, per-task responses remain unchanged. Stores that do not support serial execution ignore the new field and keep existing behavior.

Related PRs:
- TiDB: https://github.com/pingcap/tidb/pull/69686
- TiKV: https://github.com/tikv/tikv/pull/19854

Ref: pingcap/tidb#67449.

## Summary by CodeRabbit

* **New Features**
  * Clients can opt into merging data from batched task results into the main response.
  * Batched operations can return a combination of merged and separate task results.
  * Responses identify successfully merged task results and omit separate data for those results.
  * Merging is handled atomically, preventing partial merged results when an operation fails.
  * Clients can request serial execution of the primary task and batched tasks when needed; serial execution does not guarantee task order.


0xPoe (commented):
🔢 Self-check (PR reviewed by myself and ready for feedback)

- [x] Code compiles successfully

- [x] No AI-generated elegant nonsense in PR.

- [x] Comments added where necessary

- [x] PR title and description updated

- [x] Documentation PR created (or confirmed not needed)

- [x] PR size is reasonable

mjonss (approved):
LGTM, maybe the longest comments, but OK for me :)

terry1purcell (approved):

qw4990 (approved):

proto/coprocessor.proto
```diff
@@ -72,6 +72,33 @@ message Request {
...
+  //
+  // For every merged task, the store still adds a `StoreBatchTaskResponse`
+  // with `data_merged_into_response` set. The store may return some or all task
+  // results separately even when this field is set, for example when a task
+  // fails, when its result is not mergeable, or when the store predates this
+  // field, so the client must handle both merged and per-task results.
+  //
+  // This field only negotiates the response shape. It does not affect task
+  // scheduling and may be set independently of `execute_batch_tasks_serially`.
+  bool allow_batch_task_data_merge = 18;
```
cfzjywxk, line 91: How about abstract these fields into a standalone struct like
```
message AnalyzeBatchInfo {
   bool analyze_allow_batch_task_data_merge_enabled
   bool analyze_execute_batch_tasks_serially
}
```
and put it into the
```
message Request {
    AnalyzeBatchInfo
}
```

0xPoe: This is not only for analyze; it is designed for general purposes.

0xPoe: Batch request feature itself already exists for a long time. So there is no good reason in only tying it to analyze here.

0xPoe: The implementation cost on the TiKV side is similar whether we do it only for analyze or make it general.

cfzjywxk (approved):

0xPoe:
Just rebased onto the latest master. Nothing changed.

PR #1502, opened 2026-07-20
kvrpcpb: add read-pool task details

lcwangchao:
ref https://github.com/tidbcloud/cloud-storage-engine/issues/5480

## What changed

- Add an optional `PoolTaskDetails` field to `ExecDetailsV2`.
- Expose per-task dispatch, poll, wall-time, queue-wait, wake-wait, fair-queue, and poll CPU metrics.
- Regenerate the Go protobuf binding and update `scripts/proto.lock`.

## Why

Cloud Storage Engine now collects request-level read-pool scheduling and execution metrics. Returning these metrics in `ExecDetailsV2` lets clients correlate request latency with read-pool queueing and execution behavior.

Companion implementation: https://github.com/tidbcloud/cloud-storage-engine/pull/5646

This is an additive protobuf change: existing field numbers and wire behavior remain unchanged when the new field is absent.

## Verification

- `make proto-fmt-check`
- `make check`
- `make go`
- `make rust` was attempted locally, but the macOS toolchain is blocked in existing `raft-proto`/`libz-sys` build scripts; the repository's Ubuntu Rust CI will provide the authoritative Rust build result.


cfzjywxk (approved):

nolouch (commented):
Is throttling included in this pr? Can we distinguish the wait time caused by throttling from that caused by limit or fair scheduling?

lcwangchao:
> Is throttling included in this pr? Can we distinguish the wait time caused by throttling from that caused by limit or fair scheduling?

@nolouch  Not yet, mainly for two reasons:
The definition of “wait time caused by throttling” is not clear yet.
If it includes all throttling intervals between enqueue and dequeue, we would need to accumulate fragmented throttle windows for each task, which is complex.

proto/kvrpcpb.proto
```diff
@@ -1341,6 +1341,64 @@ message ExecDetailsV2 {
 
   // RU (Request Unit) consumption details.
   RUV2 ru_v2 = 5;
+
+  // Scheduling and execution details for the request's read-pool task.
+  // Available when read-pool task tracking is enabled in TiKV.
+  PoolTaskDetails read_pool_task_details = 6;
```
niubell, line 1347: I noticed there are reserved fields in ExecDetails; is it better to put it in the ExecDetails field? There will be more ExecDetailsVxx in the future; a simple and safe way, but not so clean.

proto/kvrpcpb.proto
```diff
@@ -1341,6 +1341,64 @@ message ExecDetailsV2 {
...
+}
+
+// Scheduling and execution details collected across all polls of one task running in a pool.
```
nolouch, line 1350: Can you add the more comments to explain the timing model, like:


  ```text
  The timing model is:
                             total_wall_nanos
      <--------------------------------------------------------------------->

      submitted      worker starts       async event       worker starts   done
          │                │               completes             │          │
          ▼                ▼                   ▼                 ▼          ▼
      ┌──────────────┬─────────────────┬─────────────────┬──────────────┬──────┐
      │  Queue Wait  │ Poll Execution  │ Async/Wake Wait │  Queue Wait  │ Poll │
      │              │                 │                 │              │ Exec │
      │ task is      │ Future::poll    │ Future is       │ task is      │      │
      │ runnable,    │ returns Pending │ Pending, waiting│ runnable,    │Ready │
      │ waiting for  │                 │ for I/O, timer, │ waiting for  │      │
      │ a worker     │                 │ lock, etc.      │ a worker     │      │
      └──────────────┴─────────────────┴─────────────────┴──────────────┴──────┘
             │                │                  │
             │                │                  └─ total/max/min_wake_wait_nanos
             │                │
             │                ├─ poll_wall_nanos
             │                └─ poll_cpu_nanos
             │
             ├─ total/max/min_queue_wait_nanos
             └─ fair_queue_waited_task_slices
  ```

lcwangchao: done, added this to the comment

JmPotato (approved):

PR #1516, opened 2026-08-06
pdpb: expose global barriers in GetGCState

### LINKED ISSUE #1517: pdpb: expose global GC barriers in GetGCState
## Problem

`GetGCState` currently returns one keyspace's transaction safe point, GC safe point, and keyspace-local GC barriers. However, a global GC barrier applies to every keyspace and may be the actual constraint preventing the selected keyspace from advancing its safe points.

The existing response cannot tell callers which global barriers may be limiting that keyspace. Using `GetAllKeyspacesGCStates` provides the information, but it unnecessarily returns every keyspace's GC state when the caller only needs to diagnose one keyspace.

## Proposed change

Extend the `pdpb.GetGCState` request and response schema so callers can explicitly request global GC barriers:

- Add an opt-in `include_global_gc_barriers` field to `GetGCStateRequest`.
- Keep the request field disabled by default so existing requests retain their current behavior.
- Add a `GlobalGCBarriersInfo` wrapper to `GetGCStateResponse`.
- Use wrapper-message presence to distinguish an unavailable result from a supported request that returned an empty barrier collection.
- Regenerate the protobuf bindings and schema lock.
- Add wire-compatibility coverage for old and new request and response messages.

The compatibility semantics should be:

- Old clients do not request the new information and ignore the new response field.
- New clients talking to old servers receive an absent wrapper and treat global barrier information as unavailable.
- A present-empty wrapper means the request was supported and fulfilled, but no global barriers were returned.
- A present non-empty wrapper carries the returned global barrier records.

## Related changes

- PD feature request: [tikv/pd#11116](https://github.com/tikv/pd/issues/11116)
- PD implementation: [tikv/pd#11117](https://github.com/tikv/pd/pull/11117)
- kvproto implementation: [pingcap/kvproto#1516](https://github.com/pingcap/kvproto/pull/1516)

wfxr:
Issue Number: close #1517

## What is changed?

- Add an opt-in `include_global_gc_barriers` field to `GetGCStateRequest`.
- Add a presence-preserving `GlobalGCBarriersInfo` wrapper to `GetGCStateResponse`.
- Regenerate the Go protobuf bindings and update the protobuf schema lock.
- Add a wire-compatibility test covering old and new request and response messages, including absent, present-empty, and populated global barrier results.

## Why?

PD's `GetGCState` API can return one keyspace's safe points and local GC barriers, but it cannot identify global GC barriers that may be limiting that keyspace. Fetching `GetAllKeyspacesGCStates` provides the information but is unnecessarily expensive for single-keyspace diagnostics.

This schema extension supports [tikv/pd#11116](https://github.com/tikv/pd/issues/11116) and [tikv/pd#11117](https://github.com/tikv/pd/pull/11117).

## Compatibility

- The new request field defaults to `false`, so existing clients and default requests retain the current behavior.
- Old clients ignore the new response field.
- New clients talking to old servers receive an absent wrapper and can treat the global barrier information as unavailable.
- A present-empty wrapper distinguishes a supported request with no stored global barriers from an unavailable result.

## Tests

- `make check`
- `go test ./...`
- `go test -count=1 ./pkg/pdpb -run TestGetGCStateGlobalBarriersWireCompatibility`

## Summary by CodeRabbit

* **New Features**
  * Added support for requesting global garbage-collection barrier information.
  * Responses can now indicate whether global barrier data was provided, including when no barriers are present.
* **Compatibility**
  * Added wire-format compatibility coverage to ensure older and newer clients can exchange global barrier data reliably.


rleungx:
Please also create an issue for this PR.

proto/pdpb.proto
```diff
@@ -1225,6 +1225,13 @@ message GetGCStateRequest {
 
   KeyspaceScope keyspace_scope = 2;
   bool exclude_gc_barriers = 3;
+  // Include all stored global GC barriers in the response.
+  //
+  // This uses an include flag, unlike exclude_gc_barriers, because proto3 bool
```
rleungx, line 1230: This updated field comment was not propagated to `pkg/pdpb/pdpb.pb.go`; the checked-in Go binding still contains the previous two-line comment, so the generated API documentation is stale and the next regeneration will dirty the repository.

rleungx (approved):
The rest LGTM

proto/pdpb.proto
wfxr: Fixed in 6d1e4e1. Ran make go to regenerate the Go binding, and pkg/pdpb/pdpb.pb.go now carries the updated field comment.

cfzjywxk (approved):

PR #1528, opened 2026-09-02
pdpb: add WatchGCStates streaming RPC

### LINKED ISSUE #1529: pdpb: add WatchGCStates streaming RPC
## Problem

Clients that need to react to PD GC state changes currently have to poll `GetAllKeyspacesGCStates`. Repeatedly fetching the complete state of every keyspace is inefficient for physical clusters that host many keyspaces.

The GC API refactor originally deferred a watch API. It is now needed for Starter and Essential clusters migrating from `pd-cse` to upstream PD, including a compatible replacement for the existing `WatchGCSafePointV2` behavior.

## Proposed change

Add a server-streaming `WatchGCStates` RPC to the PD service:

- Return GC state updates for all affected keyspaces through a stream.
- Send the current state of all keyspaces when the stream starts unless `skip_loading_initial` is set.
- Allow callers to set `exclude_gc_barriers` to reduce network traffic and omit events where only GC barriers changed.
- Regenerate the Go protobuf bindings and update the protobuf schema lock.

## Related changes

- PD feature request: [tikv/pd#10497](https://github.com/tikv/pd/issues/10497)
- GC API refactor: [tikv/pd#8978](https://github.com/tikv/pd/issues/8978)
- Superseded kvproto PR: [pingcap/kvproto#1440](https://github.com/pingcap/kvproto/pull/1440)
- kvproto implementation: [pingcap/kvproto#1528](https://github.com/pingcap/kvproto/pull/1528)

wfxr:
Issue Number: close #1529

Closes #1440, which this PR supersedes because its original head branch is no longer writable.

## Summary

- Add a server-streaming `WatchGCStates` RPC to PD.
- Allow clients to skip the initial state snapshot and exclude GC barrier details.
- Regenerate the Go bindings and update the proto compatibility lock.

## Related issues

- Ref: https://github.com/tikv/pd/issues/10497
- Ref: https://github.com/tikv/pd/issues/8978

## Tests

- `make go`
- `go test -count=1 ./pkg/pdpb`

## Summary by CodeRabbit

* **New Features**
  * Added a streaming API for monitoring garbage-collection state changes.
  * Streams ordered updates when keyspace garbage-collection state is added, changed, or removed.
  * Documented reconnection and event-ordering behavior for reliable monitoring.
  * Added approximate row-based key-value storage size reporting for the IA tier in region heartbeats.
* **Changes**
  * Garbage-collection barrier details are omitted from streamed effective states.
  * Removed the option to include or exclude barrier details in stream requests.


cfzjywxk (approved):

proto/pdpb.proto
```diff
@@ -1272,6 +1274,27 @@ message GetAllKeyspacesGCStatesResponse {
...
+  // amount of data transferred over network. Therefore we add this parameter to
+  // support explicitly excluding GC barriers, and ignoring events where only GC
+  // barriers are changed.
+  bool exclude_gc_barriers = 3;
+}
+
+message WatchGCStatesResponse {
+  ResponseHeader header = 1;
+
+  repeated GCState gc_states = 2;
```
rleungx: When `skip_loading_initial` is false, this response has neither a revision nor a snapshot-boundary marker. In the linked PD implementation the listener is registered before the initial scan runs asynchronously, so a live update can be queued before an older value collected by that scan; a native `WatchGCStates` client can then observe, for example, safe point 200 followed by stale safe point 100, with no way to reject the fallback.

rleungx: `GetAllKeyspacesGCStates` returns global GC barriers separately, but this watch response can only carry per-keyspace `GCState.gc_barriers`. Therefore a caller using the default `exclude_gc_barriers = false` still cannot observe global-barrier additions, removals, or expiry, so replacing polling with this stream yields an incomplete view of GC state.

wfxr: I see the concern. After looking at the intended consumers, I’m leaning toward not exposing GC barrier details through `WatchGCStates` at all.

The stream is mainly intended for GC executors to observe effective GC state changes, such as changes to `txn_safe_point` or `gc_safe_point`. GC barriers are inputs used internally by PD’s GC manager to calculate those values, and other components should not need to depend on the barrier records themselves.

For troubleshooting and diagnostics, the recently added `get gc-state` command in pd-ctl already provides the full picture, including both per-keyspace GC barriers and cluster-wide global GC barriers.

To make this scope explicit, I propose removing `exclude_gc_barriers` and documenting that `WatchGCStates` neither includes barrier details nor emits events for barrier-only changes. If a barrier change later results in an actual GC state change, that updated state will still be sent through the stream.

This would also keep the implementation simpler: the watcher would no longer need separate barrier-loading and filtering paths, or special handling for barrier-only events, for a use case we do not currently have.

Would this direction address your concern?

wfxr: You are right, and I’ve noticed the issue and plan to address it in the PD implementation as part of tikv/pd#10498.

> One possible solution is to maintain a per-watcher `dirtyDuringInitial` set. Live updates mark the keyspace dirty under `GCStateManager.mu`; before an initial batch is enqueued, initial entries for dirty keyspaces are filtered under the same mutex. This guarantees that either the initial value is enqueued before the live value, or the stale initial value is omitted.

See https://github.com/tikv/pd/pull/10498#pullrequestreview-5097472165 for details

rleungx (approved):

PR #1534, opened 2026-09-15
*: add txn protocol version declaration

### LINKED ISSUE #1535: kvproto: register txn protocol version and request declaration
ref: https://github.com/tikv/tikv/issues/20085

## Background

A TiKV store cannot tell whether a caller understands a newer transaction protocol semantics. When a feature such as shared locks changes how requests and responses are interpreted, requests from callers that do not implement it are still served best effort, and the mismatch surfaces later as wrong or mis-decoded results rather than a clear error.

## Proposal

Register a transaction protocol version in kvproto and let every transactional request declare the version it implements in `Context.txn_protocol_version`. A store compares the declaration against the range it admits and rejects an unsupported request with a structured error instead of processing it.

## Scope of the kvproto change

- `proto/kvrpcpb.proto`: `TxnProtocolVersion` registry (`TXN_VER_LEGACY = 0`, `..._INCOMPATIBLE_ERROR_HANDLING = 1`, `..._SHARED_LOCK = 2`), `Context.txn_protocol_version = 39`; `RequestOrigin` extended with TiCDC, BR and TiFlash.
- `proto/errorpb.proto`: `IncompatibleRequestReason`, `IncompatibleRequest` and `Error.incompatible_request = 23`, including the legacy `ServerIsBusy.reason = "txn_protocol_incompatible"` fallback for requests that declared version `0`.
- `proto/metapb.proto`: `TxnProtocolVersionRange` and `Store.txn_protocol_version_range = 14`, with presence kept so that a missing range stays distinguishable from an explicit `[0, 0]`.
- Regenerated Go bindings plus the `proto.lock` update.

## Out of scope

This change only defines the wire contract. TiKV admission and shared-lock fail-closed enforcement, PD updating and persisting the store range, client declaration injection and incompatibility handling, and wire compatibility tests are left to the respective repositories and follow-up PRs.

zyguan:
Issue Number: close #1535

Introduce a protocol version registry and the request declaration that lets a client tell the server which transaction protocol semantics it implements, so that requests from callers which cannot handle a feature can be rejected consistently instead of being served with best effort.

- kvrpcpb: add the TxnProtocolVersion enum and Context.txn_protocol_version (tag 39). Version numbers accumulate monotonically: 0 is legacy, 1 means the client handles the structured incompatibility error, and 2 covers the shared lock protocol. A missing field and an explicit 0 both mean legacy. Also extend RequestOrigin with TiCDC, BR and TiFlash and document that the origin is client-provided auditing metadata, not an authorization decision.
- errorpb: add IncompatibleRequest (reason, message, provided/min/max compatible version) and Error.incompatible_request (tag 23). The bounds are inclusive and the two error fields stay independent message fields. A rejection of a request that declared no version also sets ServerIsBusy.reason to the fixed "txn_protocol_incompatible" fallback for legacy clients; a declared non-zero version gets no fallback.
- metapb: add TxnProtocolVersionRange and Store.txn_protocol_version_range (tag 14). The message field keeps its presence so that an absent range stays distinguishable from an explicit [0, 0].

## Summary by CodeRabbit

- **New Features**
  - Added structured error details for requests using unsupported transaction protocol versions, including compatible version ranges and diagnostics.
  - Added transaction protocol version metadata to requests and stores for compatibility negotiation.
  - Added request-origin identifiers for TiCDC, BR, and TiFlash auditing.
  - Added transaction protocol metadata for shared-lock support.
- **Documentation**
  - Clarified request-origin auditing behavior and compatibility requirements.
  - Documented legacy fallback handling when no transaction protocol version is provided.


cfzjywxk (approved):
LGTM

proto/errorpb.proto
```diff
@@ -88,8 +88,30 @@ message EpochNotMatch {
   repeated metapb.Region current_regions = 1;
 }
 
+// Reasons for rejecting an incompatible transaction protocol request.
+enum IncompatibleRequestReason {
+  IncompatibleRequestReasonUnknown = 0;
+  // The declared version is outside the compatible range returned in the error.
+  IncompatibleRequestReasonTxnProtocolVersionOutOfRange = 1;
```
lcwangchao, line 95: Better to rename `IncompatibleRequestReasonUnsupportedTxnProtocolVersion` ?

zyguan: I slightly prefer `TxnProtocolVersionOutOfRange` here.

`UnsupportedTxnProtocolVersion` sounds like the protocol version itself is statically unsupported by the server, while compatibility here is contextual. The compatible range can be derived from the store's admission range, or narrowed by a feature-specific check. As a result, requests declaring the same version may be accepted in one case but rejected with `[min, max]` in another.

`OutOfRange` describes the actual rejection condition more precisely: the provided version is outside the compatible range returned with this particular error, without implying that the version is unsupported in general.

It also seems consistent with the current wire contract, which explicitly models compatibility as an inclusive `[min, max]` range.

lcwangchao (approved):

overvenus (approved):

hbisheng (approved):

